PR #17, opened 2020-11-12
Initial project setup

### LINKED ISSUE #1: Project Setup
### Acceptance Criteria
- Setup Go with Gin framework for development
- Setup Postgres

gutakk:
## What happened
- Initialize Go Gin project

## Insight
- Use `go run main.go` to start local server
- Use `make test` to run test
- Use `./bin/envsetup.sh` to run database container

## Proof Of Work
- Go app can be able to run and can connect to postgres
Screen Shot 2563-11-12 at 18 39 22

- Able to execute health API
Screen Shot 2563-11-12 at 18 40 30

- Test passed
Screen Shot 2563-11-12 at 18 41 04


sleepylee (approved):

config/config.go
```diff
@@ -0,0 +1,22 @@
...
+	"os"
+
+	"github.com/joho/godotenv"
+)
+
+func LoadEnv() {
+	env := os.Getenv("APP_ENV")
+
+	if "" == env {
+		env = "dev"
```
sleepylee: how do you think about having some sorts of const extraction instead of hard-coding the string? i.e:
```
const (
	DevMode   = "dev"
	ReleaseMode = "release"
	TestMode    = "test"
)

...
if "" == env { 
    env = DevMode
}

Comment (gutakk) [reply config/config.go]: Good catch. Fixed in 4acacf10b55cec9f6ddda7217482556f44f0bf98

Comment (gutakk) [reply config/config.go]: Fixed syntax error in 9060c968db1aea418cca47ce576424ce423d1bd3

File: .env [hunk 2]
```diff
@@ -0,0 +1,4 @@
+DB_HOST=localhost
```
olivierobert: The best practice is not to commit this file (need to add to gitignore) as there might be sensitive data down the road. Instead, commit `env.sample` and add instructions to rename this file `.env` in the project instructions. 

I saw further down in the config that you plan to have `.env.dev.local` or `.env.test.local`. But I would say that `.env`, `.env.sample` and `.env.test` would be more standard 💭

bin/envsetup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
olivierobert: Minor comment but having a bin folder in go project is less standard as in Rails. Since there is already a Makefile, how about moving that command to  that file too? 🤔 

```
dev-start:
   docker-compose -f docker-compose.dev.yml up -d 

test:
	go test ./...
```

go.mod
```diff
@@ -0,0 +1,21 @@
+module github.com/gutakk/go-google-scraper
```
olivierobert, line 1: Having Go modules is so much nicer than before 🤩

go.mod
```diff
@@ -0,0 +1,21 @@
...
+
+go 1.15
+
+require (
+	github.com/gin-gonic/gin v1.6.3
+	github.com/go-playground/validator/v10 v10.4.1 // indirect
```
olivierobert: Not sure about many of these dependencies (this one for playground and others like `protobuf`) 🤔 Might be good to only keep what you plan to use.

This [boilerplate repository](https://github.com/Massad/gin-boilerplate/blob/master/go.mod) does not have many of these dependencies for instance

bin/envsetup.sh
gutakk: Good catch. Fixed in 2e5a4248963b42ab8b16ea9be1322c2ee10a27b7

Makefile
```diff
@@ -0,0 +1,2 @@
+test:
+	go test ./...
```
olivierobert: One thing to take into account is to use [phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

```
.PHONY: test

test:
	go test ./... 
```

go.mod
gutakk: Agree with you. I setup gin following [Gin Gonic repo](https://github.com/gin-gonic/gin) and they suggest to use `go get -u github.com/gin-gonic/gin`

Removed unused dependencies in f1f3204c58bf3112b3851cbcfe420748f75a0169

olivierobert:
I did have to recheck this (as it has been a while I did not work on Go 😅 ) but using `-v` flag when running test provides some better output (see below the difference):

image

So you might want to use this 😁

.env
gutakk: Fixed in 117c274a104619052218a8225fce664c42eb8ad4 49fe981cc29244dffd1151c36206e9a0c7b94f3b

gutakk:
> I did have to recheck this (as it has been a while I did not work on Go 😅 ) but using `-v` flag when running test provides some better output (see below the difference):
> 
> image
> 
> So you might want to use this 😁

Looks great. Added in 72bb8c064231023c31607fd5933c5a7c9c7aa425

Makefile
gutakk: Thanks for pointing this.
Fixed in 72bb8c064231023c31607fd5933c5a7c9c7aa425

olivierobert (approved):

PR #18, opened 2020-11-13
Setup CI

### LINKED ISSUE #2: Setup CI
### Acceptance Criteria
- Setup CI for Github Actions

gutakk:
## What happened
- Solved #2
- Setup `lint` and `test` github actions workflow

## Insight
- N/A

## Proof Of Work
- Workflow passed (you can check [here](https://github.com/gutakk/go-google-scraper/actions))


main.go
```diff
@@ -11,5 +14,7 @@ func main() {
 
 	r := config.SetupRouter()
 
-	r.Run() // listen and serve on 0.0.0.0:8080 (for windows "localhost:8080")
+	if error := r.Run(); error != nil {
+		log.Fatal(fmt.Sprintf("Failed to start the server %v", error))
+	}
```
olivierobert, line 19: Good to add indeed 😅

.github/workflows/lint.yml
```diff
@@ -0,0 +1,13 @@
+name: Lint
+
+on: push
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
olivierobert: Prefer to add a line break between each step. Also, a bit of inconsistency with `test.yml` as name this step has no `name` attribute which I think is fine by the way.

olivierobert (commented):
Not sure why you opted for two workflow files as limiting is also part of testing. Maybe having one workflow file `test.yml` with two jobs `lint` and `unit-test` would work as well. Not a major issue though.

.github/workflows/lint.yml
gutakk: Fixed in d3861b92c960ab808b74e0869f5d30d9d4040346 9106033b6f5373274fc0787d66bfabfb20d0abd1

olivierobert (approved):

sleepylee (approved):

PR #19, opened 2020-11-13
Setup CD

### LINKED ISSUE #3: Setup CD
### Acceptance Criteria
- Setup CD for Heroku

gutakk:
## What happened
- Setup CD
- Setup Heroku staging

## Insight
- Deploy workflow will run on any branch (I created another [PR](https://github.com/gutakk/go-google-scraper/pull/20) for running the deploy workflow only when `Test` workflow is finish and only on development or master branch)

## Proof Of Work
- Deploy workflow completed
Screen Shot 2563-11-16 at 11 33 21

- [Health check on Heroku staging](https://go-google-scraper-staging.herokuapp.com/health)
Screen Shot 2563-11-16 at 11 35 01


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          if [[ $BRANCH_TAG = "latest" ]]
+          then
+            echo "::set-env name=HEROKU_APP::go-google-scraper"
+          else
+            echo "::set-env name=HEROKU_APP::go-google-scraper-staging"
+          fi
+
+      - name: Docker login to docker container registry
+        run: "echo $DOCKER_TOKEN | docker login $DOCKER_REGISTRY -u $DOCKER_USERNAME --password-stdin"
```
olivierobert: I recommend using [`docker/login-action`](https://github.com/docker/login-action) as it adds a nice DSL while removing some bash script (which is harder to maintain over time)

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+  db:
+    image: postgres:12.1
+    container_name: go_google_scraper_db
+    environment:
+      - POSTGRES_DB=go_google_scraper_development
+    ports:
+      - "5432:5432"
+    restart: unless-stopped
+
+  app:
```
olivierobert, line 4: Minor comment but prefer to organize services in alphabetical order and also usually placing the app at the top:

```
services:
  app:

  db:
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Docker build
+        run: |
+          docker-compose build
```
olivierobert, line 41: Tp use caching as [defined](https://github.com/gutakk/go-google-scraper/pull/19/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R17) in the Dockerfile, it's also necessary to pull already built images unless you plan to build new images each time 🤔 

```suggestion
		  docker-compose pull
          docker-compose build
```

docker-compose.yml
gutakk: Fixed in 8d00ee48d876299bc15e5d888bb7915901b6da3d

.github/workflows/deploy.yml
gutakk: Fixed in d03104d446efd67a72ce26731708bdd1bb4a7e06 6003f896dc55d7e84cfab0102df25c597b3f8aee

.github/workflows/deploy.yml
gutakk: I don't need to build new images each time. Fixed in 8d2b9647cb80cc63dd0b3d495febe0964a062007

olivierobert (approved):

db/db.go
```diff
@@ -20,6 +20,10 @@ func InitDB() {
 }
 
 func constructDsn() string {
+	if os.Getenv("APP_ENV") == "release" {
```
sleepylee, line 23: I think later we should manage this with an extra layer, somehow I felt that db init and app build type are very close to each other, and both are manageable by "APP_ENV",  but yet we are managing their classes separately 🤔  just a bit of inconvenient since the same check is duplicated and the same logic is fragmented imo. We can discuss later off this PR

sleepylee (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
+name: Deploy
+
+on: push
```
sleepylee, line 3: ohh wait 🤔  it will deploy on every `push`? 🤔  should there be a branch filter?

sleepylee (changes requested):
missing 1 context

sleepylee:
ohhh nvm, I just saw your #20 now

sleepylee (approved):

db/db.go
gutakk: Good catch @sleepylee, Should I refactor this in this PR? If yes how about something like
```go
func SetupApp() {
   env := os.Getenv("APP_ENV")

   switch env {
	case ReleaseMode:
		// does release setup (load envs, database setup and so on)
	case TestMode:
		// does test setup (load envs, database setup and so on)
	default:
		// does development setup (load envs, database setup and so on)
	}
}
```

gutakk: I think we can refactor this in another PR.

PR #20, opened 2020-11-16
Update deploy workflow to run when Test is completed and run only development and master

### LINKED ISSUE #3: Setup CD

gutakk:
## What happened
- Run deploy workflow when `Test` workflow is completed
- Run deploy workflow only on `development` and `master` branch

## Insight
This chore can be tested after merge to default (which is development) or master branch as `workflow_run` event must exist on the master or default branch. And this workflow will only run on the master or default branch [(ref)](https://github.community/t/workflow-runs-not-starting-after-previous-workflow-completes/128345/7)

## Proof Of Work
- Deploy workflow must run when `Test` is completed on development branch.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,15 @@
 name: Deploy
 
-on: push
+on:
+  workflow_run:
+    workflows:
+      - Test
+    branches:
+      - master
+      - development
+    types:
+      - completed
+  workflow_dispatch:
 
 env:
   DOCKER_IMAGE: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


olivierobert (approved):

.github/workflows/deploy.yml
```diff
@@ -1,6 +1,15 @@
 name: Deploy
 
-on: push
+on:
+  workflow_run:
+    workflows:
+      - Test
+    branches:
+      - master
+      - development
```
sleepylee, line 9: should we make this a staging deployment on `development` branch only? 🤔  production release should be on `master` separately yes 🤔

gutakk: Did you mean create a new workflow to deploy to production separately?
As I have [this step](https://github.com/gutakk/go-google-scraper/pull/20/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R35-R42) to set the HEROKU_APP by branch

What do you think?

sleepylee: yes I think production deployment (on `master`) should have a separate flow and condition from the normal staging deployment 🤔  do you plan to have another workflow separately then? 🤔  if I understand correctly then we may keep this one as-is to run staging deploy on these 2 cases 🙏

sleepylee (approved):

PR #22, opened 2020-11-16
As a user, I can register

### LINKED ISSUE #4: As a user, I can register
### Acceptance Criteria
- Implement registration feature
- Include unit test

gutakk:
## What happened
- [x] Created register handler including unit test
- [x] Created register page

## Insight
- [x] Used `parceljs` as a bundler
- [x] Used `tailwindcss` as a css framework
- [x] Used `gorm` as an ORM
- [x] Setup test database and `setup/teardown` for the test case

## Proof Of Work
- Register page
Screen Shot 2563-11-19 at 17 57 30

- Register successfully (redirect to home page)
Screen Shot 2563-11-19 at 17 59 30

```
 id |          created_at           |          updated_at           | deleted_at |      email      |                           password                           
----+-------------------------------+-------------------------------+------------+-----------------+--------------------------------------------------------------
 1  | 2020-11-19 10:59:09.193902+00 | 2020-11-19 10:59:09.193902+00 |            | dev@nimblehq.co | $2a$10$XQqft0mkhe9Vrlsk5bVPLO2QKYnhcZwOguRLltS961WAyBf6GJ1p6
```

- Registration errors
Screen Shot 2563-11-19 at 18 02 47
Screen Shot 2563-11-19 at 18 04 59
Screen Shot 2563-11-19 at 18 05 13
Screen Shot 2563-11-19 at 18 05 52
Screen Shot 2563-11-19 at 18 06 57
Screen Shot 2563-11-19 at 18 06 28
Screen Shot 2563-11-19 at 18 06 42


assets/stylesheets/base/_button.scss
```diff
@@ -0,0 +1,13 @@
+button {
+  @apply relative w-full text-center py-2 border border-transparent
```
olivierobert, line 2: Good call to use `@apply`. This is the right way to go 💪

assets/stylesheets/base/a.scss
```diff
@@ -0,0 +1,3 @@
+a {
```
olivierobert: Prefer grouping styles for headings, list and links into a file named `typography.scss` or `text.scss`. Naming a file after an HTML tag is a big No.

assets/stylesheets/components/_alert-danger.scss
```diff
@@ -0,0 +1,3 @@
+.alert-danger {
```
olivierobert: Since we use BEM, the component should be `_alert.scss` and in the file, there should be a basic structure like the following:

```scss
.alert {
  // shared styles for all alert types

  &--success {
    // for success messages
  }

  &--danger {
    // for errors
  }
}

```

assets/stylesheets/layout/_authentication.scss
```diff
@@ -0,0 +1,20 @@
+.auth-main {
```
olivierobert: This style now global and not namespaced to a specific layout which is not what is intended here. Layout stylesheets are quite particular as it requires to a class name on the root element of the HTML page which is the HTML tag itself:

```html
<!DOCTYPE html>
<html class="layout-authentication">
  <head>
```

Then the class name for the layout must be used to namespace the styles:

```
.layout-authentication {
  .auth-main {

  }
}
```

assets/stylesheets/layout/_authentication.scss
```diff
@@ -0,0 +1,20 @@
...
+  &__content--small {
+    @apply text-center text-sm mt-4;
+  }
+
+  &__image {
+    @apply mx-auto;
+    height: 6rem;
+  }
+
+  &__title {
```
olivierobert: Not sure if this title must be specific to the registration page. Having default styles for headings like `<h2>` in `/base` would be better.

assets/stylesheets/layout/_authentication.scss
```diff
@@ -0,0 +1,20 @@
+.auth-main {
+  @apply min-h-screen flex justify-center pt-20;
+
+  &__content {
+    width: 26rem;
+  }
+
+  &__content--small {
```
olivierobert: There is actually an HTML tag for tag `<small>`. If not then a `<p>` tag would be better than a `<div>` to wrap text only content.

controllers/register.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (r *RegisterController) register(c *gin.Context) {
+	form := &RegisterForm{}
+
+	if err := c.ShouldBind(form); err != nil {
+		for _, fieldErr := range err.(validator.ValidationErrors) {
+			c.HTML(http.StatusBadRequest, "register.html", gin.H{
+				"title": "Register",
+				"error": errorHandler.ValidationErrorToText(fieldErr),
```
olivierobert: The form values (only the email) should be resent to the template as for now on each error the form is totally reset 😢

controllers/register.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if result := r.DB.Create(&models.User{Email: form.Email, Password: string(encryptedPassword)}); result.Error != nil {
+		c.HTML(http.StatusBadRequest, "register.html", gin.H{
+			"title": "Register",
+			"error": errorHandler.DatabaseErrorToText(result.Error),
+		})
+		return
+	}
+
+	session.Set(c, "status", "Register successfully")
```
olivierobert: Should this use `session.AddFlash` instead? There could be an helper like the following:

```go
func AddFlash(c *gin.Context,  value interface{}) {
  session := sessions.Default(c)
  session.AddFlash(value)
  session.Save()
}
```

This would remove the need to clear this session value in the HomeController later:

```go
"status": session.GetAndDelete(c, "status"),
```

The the flashes can be rendered with:

```go
{{ range $flash := .flash }}
<div class="alert alert--danger">{{ $flash }}</div>
{{ end }}
```

controllers/register.go
```diff
@@ -0,0 +1,61 @@
...
+				"title": "Register",
+				"error": errorHandler.ValidationErrorToText(fieldErr),
+			})
+			return
+		}
+	}
+
+	encryptedPassword, _ := bcrypt.GenerateFromPassword([]byte(form.Password), bcrypt.DefaultCost)
+
+	if result := r.DB.Create(&models.User{Email: form.Email, Password: string(encryptedPassword)}); result.Error != nil {
```
olivierobert: To keep the controllers lean and not to mix business logic, it would be better to the password creation and database process to the model `models/user.go`.

You can check this [example](https://github.com/gothinkster/golang-gin-realworld-example-app/blob/master/users/models.go) 

https://github.com/gothinkster/golang-gin-realworld-example-app/blob/d5d69550e66cdc5d671a8b704ea20e457917e3f8/users/models.go#L54-L63

It would remove the need to import `gorm` and `bcrypt` dependencies in controllers. So controllers would just need to use the proper model interface to create a user while the business logic to create the actual user would be in the models. 

Also, `RegisterForm` should still be part of the controllers as it's part of the controller job to validate user input.

controllers/register_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	response := tests.PerformRequest(s.engine, "POST", "/register", s.headers, s.formData)
+	p, err := ioutil.ReadAll(response.Body)
+	pageError := err == nil && strings.Index(string(p), "Password not match") > 0
+
+	assert.Equal(s.T(), http.StatusBadRequest, response.Code)
+	assert.Equal(s.T(), true, pageError)
+}
+
+func (s *DBTestSuite) TestRegisterWithPasswordNotReachMinLength() {
```
olivierobert: How about `TestRegisterWithTooShortPassword`?

controllers/register_test.go
```diff
@@ -0,0 +1,131 @@
...
+	s.formData.Set("confirm-password", "123456")
+}
+
+func (s *DBTestSuite) TearDownTest() {
+	s.DB.Exec("DELETE FROM users")
+}
+
+func TestDBTestSuite(t *testing.T) {
+	suite.Run(t, new(DBTestSuite))
+}
```
olivierobert, line 72: We don't have conventions on Go yet. But this way of using `TestDBTestSuite` with separate methods for each test waybetter than nesting all tests under one method:

```go
func Test(t *testing.T) {
  t.Run("RegisterWithValidParameters", func(t *testing.T) {

  }

  t.Run("RegisterWithBlankEmai", func(t *testing.T) {

  }
}
```

Looks good to me on the structure. I am a bit concerned about the sharing of `formData` for all tests. I am wondering if it might just be better to define this in each test instead of reusing the same data object and then altering it for each test. It would create just some duplication but it would isolate each test. 

Also, in the upcoming PRs, you might want to add [Faker](https://github.com/bxcodec/faker) to generate test data.

helpers/error_handler/error_message.go
```diff
@@ -0,0 +1,32 @@
...
+		return "Password not match"
+	}
+	return fmt.Sprintf("%s is not valid", err.Field())
+}
+
+func DatabaseErrorToText(err error) string {
+	pgErr := err.(*pgconn.PgError)
+
+	switch pgErr.Code {
+	case "23505":
```
olivierobert, line 36: Really. Is that the only way to do that 😳 ? Also in another context (e.g. keyword search), it would require another error message. So that seems kind of brittle. Not a blocker now but I think this would need to be revisited.

helpers/error_handler/error_message.go
```diff
@@ -0,0 +1,32 @@
...
+func ValidationErrorToText(err validator.FieldError) string {
+	switch err.Tag() {
+	case "min":
+		return fmt.Sprintf("%s must be longer than %s", err.Field(), err.Param())
+	case "required":
+		return fmt.Sprintf("%s is required", err.Field())
+	case "email":
+		return "Invalid email format"
+	case "eqfield":
+		return "Password not match"
```
olivierobert: English:

```suggestion
		return "Passwords do not match"
```

templates/home.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
+    <link type="text/css" rel="stylesheet" href="dist/index.css" />
+    <title>{{ .title }}</title>
+  </head>
+
+  <body>
```
olivierobert: An actual layout should be defined in maybe `views/layouts/default.html` so that the `head`, `body` can be shared. In `templates/home`, there should only the the HTML for that page:

```
<h2>{{ .status }}</h2>
```

Then another layout can be created for authentication in `views/layouts/authentication.html` if needed. But maybe one layout might be enough 💡 

> As you can see, I am also proposing to group both layouts and template under one directory called views/

helpers/error_handler/error_message.go
```diff
@@ -0,0 +1,32 @@
+package helpers
```
olivierobert, line 1: Why creating a subdirectory for these helpers ant hot just having:

```
├── templates
│   ├── error_handler.go
│   └── session.go
```

olivierobert (commented):
The UI looks neat 👍 I struggled a bit to run the project as the README  has not been updated. So I recommend the following:

- Add instructions to create `.env.dev` and `.env.test` based on `.env.example`

- `.env.example` could be updated with default values:

```
DB_HOST=localhost
DB_PORT=5432
DB_NAME={{your_db_name}}
DB_USERNAME=postgres
DB_PASSWORD=postgres
```

Only the DB_NAME would vary between environments.

- Add instructions to run `npm run dev` in development OR wrap up all the dev scripts (Docker + JS + Go) into one Makefile script `make dev-start` 💡 In Ruby projects, we use [foreman](https://github.com/ddollar/foreman) with a [Procfile.dev](https://github.com/nimblehq/rails-templates/blob/develop/Procfile.dev) file while in Go project we could use [forego](https://github.com/ddollar/forego) to have nice terminal colors and process separations. Note there are [a few other alternatives](https://github.com/ddollar/foreman#ports).

helpers/error_handler/error_message.go
gutakk: Because I want to use alias package importing, something like this
```go
import error_handler "github.com/gutakk/go-google-scraper/helpers/error_handler"
error_handler.ValidationErrorToText(...)

// instead of
import "github.com/gutakk/go-google-scraper/helpers"
helpers.ValidationErrorToText(...)
```
or
```go
import session "github.com/gutakk/go-google-scraper/helpers/session"
session.Set(...)

// instead of
import "github.com/gutakk/go-google-scraper/helpers"
helpers.Set(...)
```

IMO, this will cause confusion if we don't use alias. What do you think?

helpers/error_handler/error_message.go
gutakk: Good catch. Fixed in 3eb18d18f16a21a9401e79108a6114fd2449d91e d18005ec84ed0ee5862e6b400e43bb6d33aebd87

helpers/error_handler/error_message.go
gutakk: As I'm using `GORM`. This is the provided [error list](https://github.com/go-gorm/gorm/blob/master/errors.go) and I also found that GORM doesn't support another error messages from [this issue](https://github.com/go-gorm/gorm/issues/2903)

I'm not sure if there are another good solutions to solve this but I can't find one at the moment.

controllers/register_test.go
gutakk: Agree! Fixed in 68eb572039a8be57f8027e623367a8750d172ec8

controllers/register_test.go
gutakk: > separate methods for each test way better than nesting all tests under one method

Could you please explain more about this?

helpers/error_handler/error_message.go
olivierobert: I also don't know a better way but maybe adding a `TODO` or `FIXME` can help to track this improvement to do later.

helpers/error_handler/error_message.go
olivierobert: I agree that using aliases is better in the context of controllers. Since in Go, the import is for whole packages then using a directory is needed.

controllers/register_test.go
olivierobert: Here, I am saying that the structure with standalone methods (as you did):

```go
func (s *DBTestSuite) TestRegisterWithValidParameters() {}

func (s *DBTestSuite) TestRegisterWithBlankEmail() {}
```

Is better than:

```go
func Test(t *testing.T) {
  t.Run("RegisterWithValidParameters", func(t *testing.T) {

  }

  t.Run("RegisterWithBlankEmai", func(t *testing.T) {

  }
}
```

So that's a good way to move forward with this.

assets/stylesheets/layout/_authentication.scss
gutakk: Fixed in 427ba9575708b43171ffb66586008cd9e5b23d64

helpers/error_handler/error_message.go
gutakk: Added `TODO` in 6d5025635c2d0d402ec2a361f310feed70de456b

assets/stylesheets/base/a.scss
gutakk: My bad, I didn't carefully take a look at compass
Fixed in c885eaa8d011468468ce2a933ac408f784fb4ec3

assets/stylesheets/components/_alert-danger.scss
gutakk: Fixed in 3d7c42b107767a54b626f310cf2a97fa7969892a

assets/stylesheets/layout/_authentication.scss
gutakk: Fixed in 3e9ff85ca4a2dc21e0cb2d45b83ff57f1cae7a7a

gutakk:
> The UI looks neat 👍 I struggled a bit to run the project as the README has not been updated. So I recommend the following:
> 
> * Add instructions to create `.env.dev` and `.env.test` based on `.env.example`
> * `.env.example` could be updated with default values:
> 
> ```
> DB_HOST=localhost
> DB_PORT=5432
> DB_NAME={{your_db_name}}
> DB_USERNAME=postgres
> DB_PASSWORD=postgres
> ```
> 
> Only the DB_NAME would vary between environments.
> 
> * Add instructions to run `npm run dev` in development OR wrap up all the dev scripts (Docker + JS + Go) into one Makefile script `make dev-start` 💡 In Ruby projects, we use [foreman](https://github.com/ddollar/foreman) with a [Procfile.dev](https://github.com/nimblehq/rails-templates/blob/develop/Procfile.dev) file while in Go project we could use [forego](https://github.com/ddollar/forego) to have nice terminal colors and process separations. Note there are [a few other alternatives](https://github.com/ddollar/foreman#ports).

- README instruction fixed in 71edf4d75f9b5a9b7e8ede56e7882d8b7235fdc5 120fbd4c7e411c2458fcbe856bbceb58c8eb403c
- Update default value to `.env.example` fixed in e8843073e8f3a59b9351e4da11c251691669d1b7
- Wrap all the dev command in 0670ae2de27e976b269f0ef6d23a60fb280aa2bf 61e1257bb69f869c121fe0ef1aaed562c4af3a77
  - Use [`Air`](https://github.com/cosmtrek/air) as a hot reload for go app
  - Use [`forego`](https://github.com/ddollar/forego)

controllers/register.go
gutakk: Did you mean email should resent to template even the error is `email already exists`?

olivierobert: In any case of errors on any form, it's better to fill in the form with the user-submitted data so that they can understand what they did wrong. For passwords, since it's sensitive, we should not. You could send the userForm object back in the session for instance.

controllers/register.go
gutakk: Very good catch. Fixed in 596ca767173b41222e4a5ce6e0d63e98eccec7cd

controllers/register.go
gutakk: Fixed in 0d86830b364a4e4380dfea769248c28ecc326b6c

controllers/register.go
gutakk: Fixed in 4e2d7e17370093d8bdab769c3f8bd302a73f73f4

README.md
```diff
@@ -7,21 +7,32 @@
...
 * [Docker Compose](https://docs.docker.com/compose/install/)
+* [NodeJS](https://nodejs.org/en/download/package-manager/)
 
+## Create necessary dot env files
+- Create `.env` file
+- Create env dependent files (depend on your env)
+  - `.env.release`
+  - `.env.debug` (for development)
+  - `.env.test`
+- Add values from `.env.example` (for env dependent variables eg. `DB_NAME` must add to dependent env  files)
```
olivierobert: Let's add some space between each section for maintainability and follow the heading hierarchy (h2 -> h3 not h4)

```suggestion
- Add values from `.env.example` (for env dependent variables eg. `DB_NAME` must add to dependent env  files)

## Usage

### Run the Go application for development
```

olivierobert: English:

```suggestion
- Add values from `.env.example` (env dependent variables eg. `DB_NAME` must be added to each dependent env file)
```

go.mod
```diff
@@ -3,8 +3,15 @@ module github.com/gutakk/go-google-scraper
 go 1.15
 
 require (
+	github.com/cosmtrek/air v1.21.2 // indirect
```
olivierobert, line 7: Having an issue to set this up 

Screen Shot 2563-11-23 at 13 17 04

So I cannot run the app at the moment.

controllers/register.go
```diff
@@ -0,0 +1,65 @@
...
+			c.HTML(http.StatusBadRequest, "register.html", gin.H{
+				"title": "Register",
+				"error": errorHandler.ValidationErrorToText(fieldErr),
+				"email": form.Email,
+			})
+			return
+		}
+	}
+
+	hashedPassword, hashError := models.HashPassword(form.Password)
```
olivierobert: Can this method really fail? If not, I would recommend moving that logic to the models or another object (e.g. user or service object). The controller should not worry about this kind of implementation details. 

As I explained in a previous comment, for this project, I think we can just move most logic to the model. So `models.SaveUser` could also call `models.HashPassword`. On large projects, it would better to have service or use case objects e.g. `services/create_user.go` or `use_cases/registration/create_user.go` which then can handle complex business logic. What we want really in the controller is just to return success/error code and messages

olivierobert (commented):
Looking good. Let's find a way to improve views and recheck what to keep in controllers.

gutakk:
@olivierobert I added more instruction to start development server more properly in this PRs a25ad07133f0dd2a0a234337c38285701e2de9ea 4c1ed16c7a3e3bd85c6e73a96d4f992d0bc3044e

I need to add `build-dependencies` step because `air` and `forego` need a binary file in order to use their command but `go mod download` doesn't create binary file.

templates/home.html
gutakk: Fixed in 42cfcb606d6c9c215b7f6bbbac1b34a6b3ac2b33 100ed0110463202461d3da01078c677002ff687c

go.mod
gutakk: please check this [comment](https://github.com/gutakk/go-google-scraper/pull/22#issuecomment-732070303)

README.md
gutakk: Fixed in 2f18e660051fb9921c8edd14b1a4c70a615a8f6e

gutakk: Fixed in 30d1be4827d47295b8549c54aeef5c15ce040b76

controllers/register.go
gutakk: This is bcrypt `GenerateFromPassword` implementation
```go
func GenerateFromPassword(password []byte, cost int) ([]byte, error) {
	p, err := newFromPassword(password, cost)
	if err != nil {
		return nil, err
	}
	return p.Hash(), nil
}
```
it returns error but I'm not sure if this method could really fail.
What do you think?

Makefile
```diff
@@ -1,7 +1,15 @@
 .PHONY: test env-setup
 
+build-dependencies:
+	go get github.com/ddollar/forego
+	go get github.com/cosmtrek/air
```
olivierobert, line 5: Good addition in the end 👍 All good on my end now.

controllers/register.go
olivierobert: Unlikely. But your call on this to keep the error checking or not. In any case, I don't think that password hashing should be the responsibility of the controller so calling it inside `SaveUser` makes more sense. Then the method `SaveUser` could return the hashing error if it happens.

controllers/register_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (s *DBTestSuite) TearDownTest() {
+	db.GetDB().Exec("DELETE FROM users")
+}
+
+func TestDBTestSuite(t *testing.T) {
+	suite.Run(t, new(DBTestSuite))
+}
+
+func (s *DBTestSuite) TestRegisterWithValidParameters() {
```
olivierobert, line 74: Now that `saveUser` method has been moved to the model, many of those tests can be moved at the model level (right now there are not tests). So the controller could have simpler tests (1 test for valid data and 1 for invalid data), that's the advantage of moving logic to model or service objects.

templates/layouts/master.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
```
olivierobert: I forgot to raise this before so sorry for the delay (I thought I did). But there are some missing critical meta tags.

- Never omit the charset as it can break the rendering:

```
<meta charset='utf-8'>
```

Read more about it in [this classic article](https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/).

- The layout is not responsive now as it's missing this meta tag:

```
<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
```

Nor the content is just scaled down to the screen size instead of taking the whole page size:

image

The other meta tags can be omitted but these two are definitely compulsory.

templates/register.html
```diff
@@ -0,0 +1,34 @@
...
+<main class="auth-main">
+  <div class="auth-main__content">
+    <img class="auth-main__image" src="dist/images/google-go.jpg" alt="Workflow">
+
+    <h2 class="auth-main__title">
+      Register a new account
+    </h2>
+    
+    {{ if .error }}
+      <div class="alert--danger">
```
olivierobert: This class is NOT BEM and it breaks (no styles):

image


Instead, it should be:


```suggestion
      <div class="alert alert--danger">
```

Each class has a purpose:

- `.alert`: shared styles for all alerts e.g. size, dimensions, positioning
- `.alert--danger`: styles only for error messages e.g. color, background-color

olivierobert (changes requested):
Good on you to resolve the local development setup. So mostly some testing and frontend improvements left 💪 Geeting there!

gutakk:
@olivierobert 
I add more unit test as following
- [x] [ValidationErrorToText](ffb5106cc2197f3bdb48e309f836c9c5e72f6f51)
- [x] [DatabaseErrorToText](43f184d50f5d76a5cbc60ed3f128ecc61dbde9b8)

assets/stylesheets/layout/_authentication.scss
gutakk: Fixed in 34994d6660857d0769b6692bdce34d467645d649
I decided to use `layout-default` for now as I don't have an idea how keyword page should look like. If keyword page layout is not the same as this one I will add another layout in that PR.

controllers/register.go
gutakk: Fixed in 659be4d2cb42d2236ac4eb36e9ba9824943b0a69.

I didn't check hash password error because I think it's ok to let `Create` function handle the error instead.

controllers/register_test.go
gutakk: Add user model unit test in 1bfcb6ebf3caab5a5f0d2804f98549839b9655d7 295c3b970e6c47f80b1461440bd6dafe40e25892

I think we cannot remove invalid data test cases because all of the invalid data test cases are validation. IMO this is controller responsibility.

What do you think?

gutakk: Also using `faker` to generate random data for each test case

templates/layouts/master.html
gutakk: Thank you so much for the knowledge and resource.
Fixed in 421e7392c42836cc55af919268121a80a6b04ede

templates/register.html
gutakk: it was my sloppy.
Fixed in a6bb092cd493d30437f722b3f56d908fefa8dfd8

assets/stylesheets/layout/_authentication.scss
gutakk: I just change my mind to follow your suggestion as I think keyword page layout will different from authentication.
Fixed in 8e85b76230039099e863cb1ea1f119f202d60d0d and c3d02537777746b1f11d027b544f5ac14ac25516

gutakk:
@olivierobert 
Please also take a look at this [comment](https://github.com/gutakk/go-google-scraper/pull/22#discussion_r527381130)

templates/register.html
olivierobert: ~~Still having some issues with the latest code. While the class names are now correctly added, the styles still do not show:~~

image

Nevermind, I noticed afterwards that I had a local issue with the JS dependencies

Screen Shot 2563-11-25 at 13 42 26

templates/layouts/master.html
olivierobert: Let's add some padding on `main` on small screens though now that the screen is resized properly:

image

assets/stylesheets/layout/_authentication.scss
```diff
@@ -0,0 +1,22 @@
+.layout-authentication {
```
olivierobert: Minor comment on the directory name as it should be `layouts/` to be consistent with `components/`

controllers/register_test.go
olivierobert: > I think we cannot remove invalid data test cases because all of the invalid data test cases are validation. IMO this is controller responsibility.

Controllers must validate the user input indeed. That's where the struct `RegisterForm` is useful. And tests in controllers should be reserved to HTTP code checks and errors as done here. 

Right now, the controller tests seem fine but the test on the models are a bit too limited. When adding more tests to models, you will realize that some tests on the controller are somewhat redundant. But if there are as many tests on the controller as on the model then there are no issues. It's a question of efficiency: it's better to test thoroughly the business logic instead of controller flows.

assets/stylesheets/layout/_authentication.scss
gutakk: Fixed in 12f96b87968c3853494f8e01a055260ec940b3c0 20a48e36aa36579e347fdb68777a6e33aecd0c7b

templates/layouts/master.html
gutakk: Added padding in a25b2dc7b29130fe7f4bc9a31a12412f153c480f

controllers/register_test.go
gutakk: > > I think we cannot remove invalid data test cases because all of the invalid data test cases are validation. IMO this is controller responsibility.
> 
> Controllers must validate the user input indeed. That's where the struct `RegisterForm` is useful. And tests in controllers should be reserved to HTTP code checks and errors as done here.
> 
> Right now, the controller tests seem fine but the test on the models are a bit too limited. When adding more tests to models, you will realize that some tests on the controller are somewhat redundant. But if there are as many tests on the controller as on the model then there are no issues. It's a question of efficiency: it's better to test thoroughly the business logic instead of controller flows.

Update error message return type and add more user model test cases in ae0d6f9411565b4dfe1efa6e67c82417147c5873 78603a7083efc30b6932fe6c7fb0cb08b109d551

controllers/register_test.go
```diff
@@ -0,0 +1,130 @@
...
+	pageError := err == nil && strings.Index(string(p), "Password is required") > 0
+	isEmailFieldValueExist := err == nil && strings.Index(string(p), s.email) > 0
+
+	assert.Equal(s.T(), http.StatusBadRequest, response.Code)
+	assert.Equal(s.T(), true, pageError)
+	assert.Equal(s.T(), true, isEmailFieldValueExist)
+}
+
+func (s *DBTestSuite) TestRegisterWithPasswordNotMatchValidation() {
+	s.formData.Set("confirm-password", "1234567")
```
olivierobert: Minor thing but for invalid data, make it easy to reach that it's invalid without having to check the sample data created above:

```suggestion
	s.formData.Set("confirm-password", "invalid")
```

helpers/error_handler/error_message.go
```diff
@@ -0,0 +1,40 @@
...
+type FieldError interface {
+	Tag() string
+	Field() string
+	Param() string
+}
+
+func ValidationErrorMessage(err FieldError) error {
+	switch err.Tag() {
+	case "email":
+		return errors.New("Invalid email format")
```
olivierobert, line 20: Just curious, why changing to `errors.New`? What are the advantages? Seems better but I want to understand as well 😁

olivierobert (approved):
Looks good to me. Check a few minor comments then it's good to merge.

controllers/register_test.go
gutakk: Fixed in 99c62793ff03294f062aee9677b11fd77b98247d

helpers/error_handler/error_message.go
gutakk: IMO, it is more straightforward so it is much more easier to understand if the function return an error and also we can able to use Go errors package's abilities.

As I took a look at another packages. Non of those packages return string as an error as Go provides `errors` package for us. Also most of the `golang error handling best practice` articles suggest to return type `error` if the function has to return an error.

so should be something like this

```go
func ThisFunctionReturnError() error {
   return errors.New("custom error")
}

func SomeFunc() {
  err := ThisFunctionReturnError()
  
  // To use error string (errors package provide us an interface which include func Error() string {...} function 
  log.Print(err.Error())
}
```

gutakk: Also fix the test in 5093112c6388378bea2a6871005e5a6184e55651

helpers/error_handler/error_message.go
sleepylee: perhaps a const for error codes with a readable name would solve it 💡

sleepylee (approved):

helpers/error_handler/error_message.go
olivierobert: Indeed, using `errors.New` should be the preferred way. Thank you for the detailed explanation as well.

PR #23, opened 2020-11-19
As a user, I can login

### LINKED ISSUE #5: As a user, I can login
### Acceptance Criteria
- Implement login feature
- Include unit test

gutakk:
## What happened
- [x] Created Login handler with unit test
- [x] Created Login page

## Insight
- [x] Split layout for app (home page) and authentication (login and register page)
- [x] Home page display greeting message if user authenticated and display `login` and `register` link if user does not
- [x] Redirect to home page when authenticated user try to access to `login` and `register` page
- [x] Redirect to home page when login successfully
- [x] Redirect to login page with notice flash when register successfully

## Proof Of Work
- Home page when logged in
Screen Shot 2563-11-25 at 17 12 20

- Home page when not logged in
Screen Shot 2563-11-25 at 17 13 08

- Redirect back to home for authenticated user
i7ExEPIeRm

- Login successfully
DuqYl48rVz

- Login unsuccessfully
Screen Shot 2563-11-25 at 17 11 45

- Register redirect to login
UKbsOwGy2i


controllers/home_test.go
```diff
@@ -6,18 +6,58 @@ import (
...
+	// Cookie from login API Set-Cookie header
+	headers := http.Header{}
+	cookie := "mysession=MTYwNjI3ODk0NHxEdi1CQkFFQ180SUFBUkFCRUFBQUlmLUNBQUVHYzNSeWFXNW5EQWtBQjNWelpYSmZhV1FFZFdsdWRBWUVBUDRCR0E9PXxa_dKXde8j6m4z_kPgaiPYuDGHj79HxhCMNw3zIoeM6g=="
+	headers.Set("Cookie", cookie)
+
+	response := tests.PerformRequest(engine, "GET", "/", headers, nil)
+	p, err := ioutil.ReadAll(response.Body)
+	isHomePage := err == nil && strings.Index(string(p), "<title>Home</title>") > 0
+	isDisplayGreeting := err == nil && strings.Index(string(p), "<h2 class=\"greeting\">") > 0
+	isDisplayLogin := err == nil && strings.Index(string(p), "<a href=\"/login\">Login</a>") > 0
```
olivierobert: I am not sure if you plan to test for everything on each page moving forward. That seems more a good case for UI tests while unit tests for controllers would be more limited with just checking for HTTP codes and success/error cases.

controllers/register.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/gutakk/go-google-scraper/models"
 )
 
+const (
+	registerTitle = "Register"
+	registerView  = "register"
+
+	registerSuccessfully = "Register successfully"
```
olivierobert: A better way would be to use localization but I realize that it's not supported by Gin and it would require a custom [middleware](https://github.com/gin-gonic/gin/issues/1382#issuecomment-670822818). It's out of the scope of this task but it's something you could check later.

controllers/register.go
```diff
@@ -34,24 +41,24 @@ func (r *RegisterController) register(c *gin.Context) {
...
 		return
 	}
 
-	session.AddFlash(c, "Register successfully")
-	c.Redirect(http.StatusFound, "/")
+	session.AddFlash(c, registerSuccessfully)
+	c.Redirect(http.StatusFound, "/login")
+}
+
+func renderRegisterWithError(c *gin.Context, status int, errMsg string, form *RegisterForm) {
```
olivierobert: That could be moved to a helper file s it will be used on more than one controller?

olivierobert (commented):
Just an early review 😁

controllers/register.go
gutakk: Should I18n be the part of this PR? if no should I create another issue for this?

controllers/home_test.go
gutakk: Fixed in bb2c084d74eec9afbd228621a2289acde50b3515

controllers/register.go
gutakk: I implemented this way because gin.H value for each page is different (2 required values for any error html render are `title` and `errors`). For example register page has `email` while keyword or home page maybe not.
Also I don't know how to add additional field to `gin.H` something like this.
```go
// Register scenario
func RenderErrorHTML(c *gin.Context, status int, view string, title string, err error, anotherValues []interface{}) {
  data = gin.H {
    "titles": title,
    "errors": err,
    ... key value pair from anotherValues
  }
  // completed data should be like this.
  // data = gin.H {
  //  "titles": title,
  //  "errors": err,
  //  "email": ...,
  // }

   ginview.HTML(c, status, view, data)
}
```

---

Get back to current implementation.
We can create helper function for current implementation and pass the whole `gin.H` into it something like this
```go
package helpers

func RenderHTML(c *gin.Context, status int, view string, data gin.H) {
  ginview.HTML(c, status, view, data)
}
```

but I think this is redundant as we can just use `ginview.HTML(...)` directly instead wrap this in another function.

What do you think?

controllers/register.go
olivierobert: I don't think it's in the scope of this project. You could create an issue to check that out later. We should always use this certification to learn as much as possible 😁

Makefile
```diff
@@ -11,4 +11,4 @@ start-dev: env-setup
 	forego start
 
 test: env-setup
-	go test -v ./...
+	go test -v -p 1 ./...
```
olivierobert, line 14: What do the extra params `-p 1` do?

templates/register.html
```diff
@@ -22,7 +22,7 @@ <h2 class="auth-main__title">
 
 <p class="auth-main__content--small">
   Already have an account?
-  <a href="#">Login</a>
+  <a href="login">Login</a>
```
olivierobert: Missing the forward slash, no?

```suggestion
  <a href="/login">Login</a>
```

templates/login.html
```diff
@@ -0,0 +1,36 @@
...
+  <input name="email" type="email" required placeholder="Email address" value={{.email}}>
+  <input name="password" type="password" required placeholder="Password">
+
+  <button type="submit">
+    Login
+  </button>
+</form>
+
+<p class="auth-main__content--small">
+  Don't have account yet?
```
olivierobert: English:

```suggestion
  Don't have an account yet?
```

templates/login.html
```diff
@@ -0,0 +1,36 @@
...
+  </div>
+{{ end }}
+
+{{ if .errors }}
+  <div class="alert alert--danger">
+    {{ .errors }}
+  </div>
+{{ end }}
+
+
```
olivierobert, line 8: One extra blank line can be removed

config/view.go
```diff
@@ -6,7 +6,22 @@ import (
...
 
-func GoviewConfig() goview.Config {
+func AppGoviewConfig() goview.Config {
+	return goview.Config{
+		Root:      "templates",
+		Extension: ".html",
+		Master:    "layouts/application",
+		Partials:  []string{},
+		Funcs: template.FuncMap{
+			"sub": func(a, b int) int {
```
olivierobert: While this is part of the example on the repository, we don't make use of this, right? So it Can be removed 😅

templates/login.html
```diff
@@ -0,0 +1,36 @@
...
+      {{ $notice }}
+    {{ end }}
+  </div>
+{{ end }}
+
+{{ if .errors }}
+  <div class="alert alert--danger">
+    {{ .errors }}
+  </div>
+{{ end }}
```
olivierobert: For all flashes. how about defining a partial in the Goview Config

```go
return goview.Config{
  Root:      "templates",
  Extension: ".html",
  Master:    "layouts/authentication",
  Partials:  []string{"partials/flashes"}
}
```

Then use in the view as:

```go
{{template "flashes" .}}
```

You can find a detailed example on [the Goview repository](https://github.com/foolin/goview/tree/master/_examples/gin-multiple).

templates/home.html
```diff
@@ -1,8 +1,11 @@
 {{ define "content" }}
 
 <h1 class="text-primary">HELLO GOLANG</h1>
-{{ range $flash := .flashes }}
-  <h2>{{ $flash }}</h2>
+{{ if .userID  }}
```
olivierobert: It would better to have a presenter object for this to wrap view related variables together. But to keep things simple, having a better DSL with a better naming would help:

```suggestion
{{ if .authenticatedUser  }}
```

or the opposite:

```suggestion
{{ if .anomymousUser  }}
```

controllers/home_test.go
```diff
@@ -6,18 +6,46 @@ import (
...
+	}
+
+	_ = db.GetDB().AutoMigrate(&models.User{})
+
+	engine := tests.GetRouter(true)
+	new(HomeController).applyRoutes(engine)
+
+	// Cookie from login API Set-Cookie header
+	headers := http.Header{}
+	cookie := "mysession=MTYwNjI3ODk0NHxEdi1CQkFFQ180SUFBUkFCRUFBQUlmLUNBQUVHYzNSeWFXNW5EQWtBQjNWelpYSmZhV1FFZFdsdWRBWUVBUDRCR0E9PXxa_dKXde8j6m4z_kPgaiPYuDGHj79HxhCMNw3zIoeM6g=="
```
olivierobert: What is the validity of this cookie? Can this expire and break tests later?

controllers/index.go
```diff
@@ -1,13 +1,24 @@
...
+	"github.com/foolin/goview/supports/ginview"
 	"github.com/gin-gonic/gin"
+	"github.com/gutakk/go-google-scraper/config"
+	"github.com/gutakk/go-google-scraper/middlewares"
 )
 
 func CombineRoutes(engine *gin.Engine) {
-	homeController := &HomeController{}
-	homeController.applyRoutes(engine)
+	new(HomeController).applyRoutes(engine)
```
olivierobert, line 11: There should also be another group and middleware for `EnsureAuthenticationGroup`, right? This is needed to redirect to the login screen if a user is not authenticated.

controllers/register.go
olivierobert: Small English adjustment needed though:

```suggestion
	registerSuccessFlash = "Registration completed successfully"
```

middlewares/authentication.go
```diff
@@ -0,0 +1,19 @@
+package middlewares
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	session "github.com/gutakk/go-google-scraper/helpers/session"
+)
+
+func EnsureNoAuthentiction(c *gin.Context) {
```
olivierobert: Typo:

```suggestion
func EnsureNoAuthentication(c *gin.Context) {
```

models/user.go
```diff
@@ -31,3 +31,21 @@ func SaveUser(email string, password string) error {
 	}
 	return nil
 }
+
+func FindOneUserByCondition(condition interface{}) (User, error) {
```
olivierobert: Minor naming improvement suggestion:

```suggestion
func FindUserBy(condition interface{}) (User, error) {
```

I find it descriptive and nicer to use 🌟

models/user.go
```diff
@@ -31,3 +31,21 @@ func SaveUser(email string, password string) error {
...
+}
+
+func FindOneUserByID(id interface{}) (User, error) {
+	user := User{}
+	result := db.GetDB().First(&user, id)
+
+	return user, result.Error
+}
+
+func CheckPassword(hashedPassword string, password string) error {
```
olivierobert: How about:

```suggestion
func ValidatePassword(hashedPassword string, password string) error {
```

models/user_test.go
```diff
@@ -28,8 +27,23 @@ func TestHashPassword(t *testing.T) {
...
+
+	assert.Equal(t, nil, result)
+}
+
+func TestCheckPasswordWithInvalidPassword(t *testing.T) {
+	hashedPassword, _ := hashPassword("password")
+	result := CheckPassword(string(hashedPassword), "drowssap")
+
+	assert.NotEqual(t, nil, result)
+}
```
olivierobert, line 42: Why are these tests defined before `DBTestSuite` and `SetupTest`? I would be more consistent to keep all tests below these basic blocks/ Now it fees a bit messy.

controllers/user_session.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/go-playground/validator/v10"
+	errorHandler "github.com/gutakk/go-google-scraper/helpers/error_handler"
+	session "github.com/gutakk/go-google-scraper/helpers/session"
+	"github.com/gutakk/go-google-scraper/models"
+)
+
+const (
+	loginTitle = "Login"
+	loginView  = "login"
```
olivierobert: Not sure why you picked the file name `user_session.go` 🤔  It seems inspired by Rails and Devise `sessions_controller.rb` 😅  It's not consistent with `register.go`. Wouldn't `login.go` be more consistent? If you prefer to keep `session.go`, then `register.go` should be renamed to `registration.go` 💭

controllers/home.go
```diff
@@ -14,8 +16,16 @@ func (h *HomeController) applyRoutes(engine *gin.Engine) {
...
 func (h *HomeController) displayHome(c *gin.Context) {
-	c.HTML(http.StatusOK, "home", gin.H{
-		"title":   "Home",
-		"flashes": session.Flashes(c),
+	userID := session.Get(c, "user_id")
+
+	var user models.User
+	if userID != nil {
+		user, _ = models.FindOneUserByID(userID)
+	}
```
olivierobert, line 34: Since pretty much all authenticated controller would need the user data, what do you think to later (in another task) move this logic into a middleware e.g. `current_user.go`?

middlewares/authentication.go
```diff
@@ -0,0 +1,19 @@
+package middlewares
```
olivierobert: In terms of file name, wouldn't it be better to name the file closer to what the middleware name is `ensure_no_authentication.go`? That allows to a faster code discovery. If you don't having the `no` in the middle, then `ensure_guest_user.go` might sound better.

controllers/register.go
olivierobert: The first example you have shown is what I had in mind: 

```go
func renderWithError(c *gin.Context, status int, view string, title string, err error, rest ...[]string{}) {
    default_params = gin.H{
		"title":  title,
		"errors": err.Error(),
	}
    view_params = append(default_params, rest...)
	ginview.HTML(c, status, view, view_params)
}
```

Untested though 😅 but that's the idea.

Makefile
gutakk: According to `go build help`
```
-p n
		the number of programs, such as build commands or
		test binaries, that can be run in parallel.
		The default is the number of CPUs available.
```
I found that run test failed sometimes because there are more records in the database than I expected but sometimes not.

controllers/register.go
gutakk: Fixed in dfa57092de76d46cb91b1104919deebc947d1544 8a777d5413e868219e7c743c5219b875e122015b

templates/register.html
gutakk: Good catch. Fixed in 096a130cd981d6647567c8dce1f60190adc893df

templates/login.html
gutakk: Fixed in b47ee9e71f5806d2e006533e2e3dfda9f34c5066

templates/login.html
gutakk: Could you please check this again? I didn't see any extra blank line

config/view.go
gutakk: Yes, agree! Fixed in 7311140562c495d8bd27936512318ad82fc1b637

templates/login.html
gutakk: I don't know why I got 500 when I use Partials config in the goview as application didn't log anything. I did following way instead. What do you think?

fcca86412e5f03b6f3270caf678ea5a0d74df606 0e59f3f628e171594751770ed13612c16138ceb2

controllers/index.go
gutakk: I design to let user able to access home page without authenticated but yes we should have that group when user try to access keyword page but I think we should done that in keyword related PR.

What do you think?

templates/home.html
gutakk: Sounds better. Fixed in ef89d938577fe4c4c0af4346ea3496707cb17b55

controllers/home_test.go
gutakk: Nice catch. I'm not really sure about this as the whole cookie look like this
```
mysession=MTYwNjQ0OTkxN3xEdi1CQkFFQ180SUFBUkFCRUFBQUlmLUNBQUVHYzNSeWFXNW5EQWtBQjNWelpYSmZhV1FFZFdsdWRBWUVBUDRFS2c9PXxsvz7BbGtpuEyUB7G0IUTjxgBAEkBNd10pBS1NGZEtOA==; Path=/; Expires=Sun, 27 Dec 2020 04:05:17 GMT; Max-Age=2592000
```

Not sure if the cookie is valid when I don't set an `Expires`.
Do you have any idea?

middlewares/authentication.go
gutakk: Fixed in 1ced9555950fdcc901e965496eddad3251270cee

models/user.go
gutakk: Sounds better. Fixed in 752d3fca9f92953e4ecb829dc7104e2a353a218a

models/user.go
gutakk: Fixed in 6872221091bc5b812877a3b984c61f969e14acfb

models/user_test.go
gutakk: Because all the defined before `DBTestSuite` test cases are not involve to database. I think it is better to keep the test cases that only involve to database define in `DBTestSuite` so `SetupTest` won't setup unnecessary stuffs to test cases that don't use them.

What do you think?

controllers/user_session.go
gutakk: Yes I thought about rails when I did this 😅. I decided to change from `user_session.go` to `login.go`
Fixed in 95fd1731700d3159f4f4167625f0383e186d0e43

controllers/home.go
gutakk: Indeed, I will move this into a middleware in another task

middlewares/authentication.go
gutakk: In the first place, I thought to put all the authentication related middleware in the `authentication.go` such as `ensure_no_authentication`, `ensure_authentication` and so on.

But I agree with you if we named the file closer to the middleware name allows a faster code discovery so I fixed in f0ef5e0023c462ed2188d8b13e40dcb04f5419a3

controllers/register.go
gutakk: `append` won't work with `gin.H` because it is not slice.

I implemented this way instead d6739d5fc316f5743612313078207bb0c9d8e2c4.

What do you think?

gutakk: Applied above solution to home and login in 458cd93353f3366cfc2bc03932583c408d577b8e

Makefile
olivierobert: Looks like some race conditions. That fix is more like a band-aid 🩹 as basically it disables running tests in parallel. So it might have to do with the test set up 🤔 In controller tests for instance since the `SetupTest()` is shared for all tests and each test modifies the test data, there is no real isolation between each test. So I belive this is an area to research more on.

templates/login.html
olivierobert: Looks fine now 😁

models/user_test.go
olivierobert: > Because all the defined before DBTestSuite test cases are not involve to database. 

That kind of makes sense but I am not sure in what order Gin execute the tests 🤔 So does placing them before the DBTestSuite really affect the way tests are run or not?

controllers/user_session.go
olivierobert: Allgood, that's what I thought also 😁

controllers/home_test.go
olivierobert: Likely since the `Expires` header is not set, this should work all the time. By the way, even in the application, the expiration is not defined so maybe adding the Options for MaxAge would be a good addition:

```go
func Set(c *gin.Context, key string, value interface{}) {
    session := sessions.Default(c)
    session.Options(sessions.Options{
        MaxAge: 3600*24,  // 24 hours
    })
	session.Set(key, value)
	_ = session.Save()
}
```

Also in terms of key naming `mysession` could be changed to the application name 💭

controllers/index.go
olivierobert: It depends what will be on the home screen 😁 I assumed it would be the list of keywords/dashboard. So I leave that to you to decide if you plan to add this later.

controllers/register.go
olivierobert: It's an improvement to keep things DRY. I am not usually a fan of using `data` as a variable name. Phoenix (Elixir) calls this `assigns` and Rails calls this `locals` FYI. `data` works but it's not descriptive at all 🧐 

> append won't work with gin.H because it is not slice.

I hoped that `gin.H` was behaving like a slice 🙃 But thanks for pointing it out. I wished Go objects were implementing the primitives data like in Ruby 🙏  Duck typing FTW

olivierobert (approved):

Makefile
gutakk: Noted

controllers/home_test.go
gutakk: Update session name in 4b067f4557366e5a1d29960505c4f068545b3c87
Set max age for session in bf2c595ab42740c4fc30dc3cb09e65f5ce451b11

models/user_test.go
gutakk: > That kind of makes sense but I am not sure in what order Gin execute the tests 🤔 So does placing them before the DBTestSuite really affect the way tests are run or not?

I cannot find any resource for this either.

I just group them to make it more easier to read.

helpers/session/session.go
```diff
@@ -18,3 +18,18 @@ func Flashes(c *gin.Context) []interface{} {
...
+
+func Get(c *gin.Context, key string) interface{} {
+	session := sessions.Default(c)
+	return session.Get(key)
+}
+
+func Set(c *gin.Context, key string, value interface{}) {
+	session := sessions.Default(c)
+	session.Options(sessions.Options{
+		MaxAge: 3600 * 24, // 24 hours
```
sleepylee: minor, but I'm not a big fan of this hardcoded value way 🙇  perhaps you can declare a const?

`const sessionTimeoutInSec = 24 * 60 * 60`

models/user.go
```diff
@@ -31,3 +31,21 @@ func SaveUser(email string, password string) error {
 	}
 	return nil
 }
+
+func FindOneUserBy(condition interface{}) (User, error) {
```
sleepylee: not sure if we need `One` 😅  if we are writing `User` in the singular that should indicate a single User in response already

models/user.go
```diff
@@ -31,3 +31,21 @@ func SaveUser(email string, password string) error {
...
+	user := User{}
+	result := db.GetDB().Where(condition).First(&user)
+
+	return user, result.Error
+}
+
+func FindOneUserByID(id interface{}) (User, error) {
```
sleepylee: same here

models/user_test.go
```diff
@@ -74,17 +92,47 @@ func (s *DBTestSuite) TestSaveUserWithDuplicateEmail() {
 }
 
 func (s *DBTestSuite) TestSaveUserWithEmptyStringEmail() {
+	db.GetDB().Exec("DELETE FROM users")
```
sleepylee, line 95: out of curiosity,  I saw that we already had `TearDown` that does the same thing here, do we really need to explicitly call this again? 🤔

sleepylee (commented):
done my review 🙇

models/user_test.go
olivierobert: Not sure about making this as a convention though. I would prefer to keep the test setup at the top and then have all tests below it. 

```
# Now

setup
some tests
DB setup
more tests

# More standard

setup
DB setup
all tests

```

The tests which don't use the DB would not take the DB param which I think is readable enough IMO. It's a matter to all agree on obviously.

models/user_test.go
olivierobert: Good catch, I did not notice this and I am also wondering why it's needed.

gutakk: Because `SetupTest` always insert 1 user to database and I have more test cases that required to have user in db before run test more than not required.

I decided to delete before run the test because I want to make sure database is clean and I can assert `row in database = 0`
```go
assert.Equal(s.T(), int64(0), result.RowsAffected)
```

What do you guys think?

helpers/session/session.go
gutakk: Fixed in d1b2b811f023220b13f5fe50ac93bac97c996c80

models/user.go
gutakk: Make sense. Fixed in 5ce542d99b7688e89218339e206436fb78659361 1c6171adf5c87a98fda697317ead8a29783b77a5

models/user.go
gutakk: Make sense. Fixed in 5ce542d99b7688e89218339e206436fb78659361 1c6171adf5c87a98fda697317ead8a29783b77a5

sleepylee (approved):

PR #24, opened 2020-11-27
As a user, I can logout

### LINKED ISSUE #6: As a user, I can logout
### Acceptance Criteria
- Implement logout feature
- Include unit test

gutakk:
## What happened
- [x] Created Logout page
- [x] Created Logout handler with unit test

## Insight
- [x] Added ensure authenticated user middleware
- [x] Added notice flash message to home page after logged out

## Proof Of Work
- Logout button at home page
Screen Shot 2563-11-27 at 17 35 18

- Logout button clicked
dvZrTA0dgl

### DIFF
--- a/controllers/logout.go
+++ b/controllers/logout.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	logoutSuccessFlash = "You've been logged out"
+	logoutSuccessFlash = "You have been logged out"
 )
 
 type LogoutController struct{}
--- a/controllers/logout.go
+++ b/controllers/logout.go
@@ -1,0 +1,16 @@
+package controllers
+
+import (
+	"github.com/gin-gonic/gin"
+	session "github.com/gutakk/go-google-scraper/helpers/session"
+)
+
+type LogoutController struct{}
+
+func (l *LogoutController) applyRoutes(engine *gin.Engine) {
+	engine.POST("/logout", l.logout)
+}
+
+func (l *LogoutController) logout(c *gin.Context) {
+	session.Delete(c, "user_id")
+}
--- a/controllers/logout.go
+++ b/controllers/logout.go
@@ -1,8 +1,15 @@
 package controllers
 
 import (
+	"net/http"
+
 	"github.com/gin-gonic/gin"
+	html "github.com/gutakk/go-google-scraper/helpers/html"
 	session "github.com/gutakk/go-google-scraper/helpers/session"
+)
+
+const (
+	logoutSuccessFlash = "You've been logged out"
 )
 
 type LogoutController struct{}
@@ -13,4 +20,7 @@ func (l *LogoutController) applyRoutes(engine *gin.Engine) {
 
 func (l *LogoutController) logout(c *gin.Context) {
 	session.Delete(c, "user_id")
+
+	session.AddFlash(c, logoutSuccessFlash, html.FlashNoticeKey)
+	c.Redirect(http.StatusFound, "/")
 }
--- a/controllers/logout.go
+++ b/controllers/logout.go
@@ -4,7 +4,6 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
-	html "github.com/gutakk/go-google-scraper/helpers/html"
 	session "github.com/gutakk/go-google-scraper/helpers/session"
 )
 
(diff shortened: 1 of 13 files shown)


controllers/logout.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	html "github.com/gutakk/go-google-scraper/helpers/html"
+	session "github.com/gutakk/go-google-scraper/helpers/session"
+)
+
+const (
+	logoutSuccessFlash = "You've been logged out"
```
olivierobert: Always prefer non-contracted English spelling:

```suggestion
	logoutSuccessFlash = "You have been logged out"
```

The contracted spelling for spoken formats only.

controllers/logout_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	assert.Equal(s.T(), http.StatusFound, response.Code)
+	assert.Equal(s.T(), "/", response.Header().Get("Location"))
+}
+
+func (s *LogoutTestSuite) TestLogoutWithGuestUser() {
+	response := tests.PerformRequest(s.engine, "POST", "/logout", nil, nil)
+
+	assert.Equal(s.T(), http.StatusFound, response.Code)
+	assert.Equal(s.T(), "/login", response.Header().Get("Location"))
```
olivierobert, line 43: I don't see how this test tell us anything as it has the same assertions as `TestLogoutWithAuthenticatedUser` 🙃  Since logout requires an authenticated user, it could throw a 401 error if it was an API but for a web application, setting a flash and redirect is better. So how about adding an assertion for the flash message?

helpers/html/render_html.go
```diff
@@ -13,17 +19,22 @@ func RenderWithError(c *gin.Context, status int, view string, title string, err
...
-func RenderWithNotice(c *gin.Context, status int, view string, title string, notices []interface{}, data map[string]interface{}) {
+func RenderWithFlash(c *gin.Context, status int, view string, title string, data map[string]interface{}) {
+	flashKey, flashValue := getFlashMessage(c)
+
 	ginview.HTML(c, status, view, gin.H{
-		"title":   title,
-		"notices": notices,
-		"data":    data,
+		"title":  title,
+		flashKey: flashValue,
```
olivierobert: This implementation is quite complex:
- It relies on `getFlashMessage`
- It requires to import `html "github.com/gutakk/go-google-scraper/helpers/html"` each time a new flash message must be added to use the constants `FlashNoticeKey` and `FlashErrorKey`

How about instead:

- Add a flash key when adding a new Flash:

Session.addFlash accepts a second parameter: https://github.com/gorilla/sessions/blob/master/sessions.go#L65

So it's possible to do:

```
session.AddFlash("This is an info message", "notice")
session.AddFlash("This is an error message", "error")
```
 
- Keeping the HTML variable to `flashes` (or `errors`?) and `notices` 

```
ginview.HTML(c, status, view, gin.H{
  "title": title,
  "errors": session.Flashes("error"),
  "notices": session.Flashes("notice")
  "data": data,
})
```

Then in the HTML template, iterate over the HTML variables for the content of each:

```
{{range .errors}}{{.}}{{end}}
{{range .notices}}{{.}}{{end}}
```

templates/home.html
```diff
@@ -1,8 +1,13 @@
 {{ define "content" }}
 
 <h1 class="text-primary">HELLO GOLANG</h1>
-{{ if .data.authenticatedUser  }}
+{{ include "partials/flashes" }}
+
+{{ if .data.authenticatedUser }}
```
olivierobert, line 6: Ideally, later, it would be great to have something like `currentUser` set in middleware as discussed before. So it would not be in `.data` ⭐

gutakk: This will be implemented in upload csv PR

controllers/logout.go
gutakk: Fixed in e599ca2117f078213bcd7b9ac559b116d9fb3278

helpers/html/render_html.go
gutakk: Fixed in 01e1b958ff94cdaf5702b9fb683bdc5c84cce769

controllers/logout_test.go
gutakk: As per we discussed in slack, added the todo comment in fac19d0124358e3173582b65d7a05b5a64f2c538

helpers/html/render_html.go
olivierobert: Looks neat 💪

olivierobert: Now after going through the code 🙈 an improvement I still can see is to have:

- `session.AddNoticeFlash(c, logoutSuccessFlash)` instead of `session.AddFlash(c, logoutSuccessFlash, "notice")`
- `session.AddErrorFlash(c, loginError)` instead of `session.AddFlash(c, logoutSuccessFlash, "error")`

This can be done in the next PR too 😁

olivierobert (approved):

sleepylee (approved):

PR #25, opened 2020-11-30
As a user, I can upload CSV file

### LINKED ISSUE #7: As a user, I can upload csv
### Acceptance Criteria
- Implement upload CSV feature
  - UI to display uploaded keywords
  - Backend to process upload task
- Only authenticated user can use this feature
- Allow only CSV file type
- Allow only 1-1000 keywords CSV file
- Include unit test

gutakk:
## What happened
- [x] Create keyword upload page

## Insight
- [x] Add current user middleware
- [x] Add keyword upload handler with unit test
  - Validate file type
  - Validate CSV length
  - Save keywords to database
- [x] Refactor test

## Proof Of Work
- Keyword upload page
Screen Shot 2563-12-02 at 16 28 41

- Access to keyword upload page without login
Screen Shot 2563-12-02 at 16 30 01

- File uploaded successfully
Screen Shot 2563-12-02 at 16 29 21

```
 id  |          created_at           |          updated_at           | deleted_at |      keyword      | user_id 
-----+-------------------------------+-------------------------------+------------+-------------------+---------
 132 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Eden Hazard       |       8
 133 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Lionel Messi      |       8
 134 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Cristiano Ronaldo |       8
 135 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Kylian Mbappe     |       8
 136 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Neymar            |       8
 137 | 2020-12-02 09:31:11.351521+00 | 2020-12-02 09:31:11.351521+00 |            | Kevin De Bruyne   |       8
(6 rows)
```

- File uploaded unsuccessfully
Screen Shot 2563-12-02 at 16 29 06


controllers/keyword.go
```diff
@@ -0,0 +1,54 @@
...
+	engine.GET("/keyword", k.displayKeyword)
+	engine.POST("/keyword", k.uploadKeyword)
+}
+
+func (k *KeywordController) displayKeyword(c *gin.Context) {
+	html.RenderWithFlash(c, http.StatusOK, keywordView, keywordTitle, nil)
+}
+
+func (k *KeywordController) uploadKeyword(c *gin.Context) {
+	file, _ := c.FormFile("file")
```
olivierobert: How about using a struct like in the registration and login? Having a struct to validate params is great even if there is only one param 💪

controllers/keyword.go
```diff
@@ -0,0 +1,54 @@
...
+	// Validate if file is CSV type
+	if err := models.ValidateFileType(file.Header["Content-Type"][0]); err != nil {
+		html.RenderWithError(c, http.StatusBadRequest, keywordView, keywordTitle, err, nil)
+	}
+
+	filename := "dist/" + filepath.Base(file.Filename)
+	_ = c.SaveUploadedFile(file, filename)
+	csvfile, _ := os.Open(filename)
+	r := csv.NewReader(csvfile)
+	record, _ := r.ReadAll()
```
olivierobert: Line 38 -> 42 should also be moved outside of the controller for the same reasons I gave you before: it's the business domain 💍 There needs to be an API to perform CSV upload so the same logic would be needed. Encapsulating it into a new object would make it reusable e.g. `keywordFile` and a method `keywordFile.upload()`.

Great to move `ValidateCSVLength` and `SaveKeywords` to the keyword model 💪

models/keyword.go
```diff
@@ -0,0 +1,59 @@
...
+		return errors.New(fileLengthError)
+	}
+	return nil
+}
+
+func SaveKeywords(record [][]string) ([]Keyword, error) {
+	var keywords = []Keyword{}
+
+	// Check if record more than 0
+	if len(record) <= 0 {
```
olivierobert: Can this be really negative?

models/keyword.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func ValidateCSVLength(row int) error {
+	if row <= 0 || row > 1000 {
+		return errors.New(fileLengthError)
+	}
+	return nil
+}
+
+func SaveKeywords(record [][]string) ([]Keyword, error) {
```
olivierobert: What will be the CSV format be like? I am not sure why it needs to be an array of array of strings 🤔

models/keyword_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func TestValidateCSVLengthWithZeroRow(t *testing.T) {
+	result := ValidateCSVLength(0)
+
+	assert.Equal(t, "CSV file must contain between 1 to 1000 keywords", result.Error())
+}
+
+func TestValidateCSVLengthWithMinusOneRow(t *testing.T) {
+	result := ValidateCSVLength(-1)
```
olivierobert: How can this case even happen 🤔 ?

models/keyword_test.go
```diff
@@ -0,0 +1,123 @@
...
+	assert.Equal(t, "File must be CSV format", result.Error())
+}
+
+func TestValidateCSVLengthWithOneRow(t *testing.T) {
+	result := ValidateCSVLength(1)
+
+	assert.Equal(t, nil, result)
+}
+
+func TestValidateCSVLengthWithOneThoudsandRow(t *testing.T) {
```
olivierobert: The min/max values might change so it would better to named some tests in a more generic way e.g. `TestValidateCSVLengthWithMaxRowAllowed`

models/keyword_test.go
```diff
@@ -0,0 +1,123 @@
...
+	suite.Run(t, new(KeywordDBTestSuite))
+}
+
+func (s *KeywordDBTestSuite) TestSaveKeywordsWithValidParams() {
+	keywords := [][]string{
+		[]string{"Hazard"},
+		[]string{"Ronaldo"},
+		[]string{"Neymar"},
+		[]string{"Messi"},
+		[]string{"Mbappe"},
```
olivierobert: Based on my previous comments, wouldn't be simpler not to have the nested arrays?

```go
[]string{"Hazard", "Ronaldo", "Neymar", "Messi", "Mbappe"}
```

In the CSV parsing, the data could be cleaned to have this format, no?

models/setup_test.go
```diff
@@ -0,0 +1,13 @@
+package models
+
+import (
+	"os"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+)
+
+func TestMain(m *testing.M) {
```
olivierobert, line 10: Why adding this file now?

olivierobert (commented):
Some early feedback 👀

models/keyword.go
gutakk: As I used `ReadAll()` function and the signature look like this
```go
func (r *Reader) ReadAll() (records [][]string, err error) {...}
```

models/setup_test.go
gutakk: Because `TestMain` can declare only once per package [(REF)](https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc)

> Since function names need to be unique in a given package, you can only define the TestMain once for each package. If your package has multiple test files, choose a logic place for your single TestMain function.

models/keyword_test.go
gutakk: As I used `ReadAll()` function and the signature look like this
```go
func (r *Reader) ReadAll() (records [][]string, err error) {...}
```

or is it better to loop through file and use `Read()` to construct slice manually instead?

models/keyword.go
gutakk: We can discuss about this in this [comment](https://github.com/gutakk/go-google-scraper/pull/25#discussion_r532476260)

controllers/keyword.go
gutakk: Fixed in 3dcfa7705031d36383aba55681a40f9a7c94ddc0

models/keyword.go
gutakk: No it doesn't. Fixed in 56662ee0b4df157d4d83680122d71b54be3a45aa

models/keyword_test.go
gutakk: Good catch. it is impossible. Fixed in b2b3d88f76e43e6cdf496f6033eac0fce9a58c05

models/keyword_test.go
gutakk: Fixed in b02cbc6b0abe3720c279b654eaecdd704f614207

controllers/keyword.go
gutakk: Fixed in 45e248a014c3d66aa52efa2092fa3bc3f97eb9c4

models/keyword_test.go
gutakk: I think it is much more better to use `Read()` to create `[]string` slice
Fixed in cd31b5299e42b4d28439db3a244e1cf1d872adae

controllers/keyword.go
olivierobert: There could also be another model for KeyworsdFile specifically for the file validation 💡

models/setup_test.go
olivierobert: Okay, so basic reason indeed 😅

controllers/keyword_test.go
```diff
@@ -0,0 +1,131 @@
...
+	user := models.User{Email: email, Password: string(hashedPassword)}
+	db.GetDB().Create(&user)
+
+	formData := url.Values{}
+	formData.Set("email", email)
+	formData.Set("password", password)
+
+	headers := http.Header{}
+	headers.Set("Content-Type", "application/x-www-form-urlencoded")
+	response := testHttp.PerformRequest(s.engine, "POST", "/login", headers, formData)
```
olivierobert: Sorry for not noticing this before, but having to manually log in for a unit tests seem awkward and not standard 🤔 Wouldn't saving a cookie as a fixture and set it manually would not work? Let me know if this is because of a limitation of Gin.

helpers/error_handler/error_message.go
```diff
@@ -44,6 +46,8 @@ func DatabaseErrorMessage(err error) error {
 	pgErr := err.(*pgconn.PgError)
 
 	switch pgErr.Code {
+	case foreignKeyErrorCode:
+		return errors.New(somethingWentWrongError)
```
olivierobert, line 50: Having so generic message is not so nice and it could create debugging issues down the road 😵 I don't have a better option now

helpers/error_handler/error_message.go
```diff
@@ -8,14 +8,16 @@ import (
...
-	minError            = "%s must be longer than %s"
-	passwordEqError     = "Passwords do not match"
-	requiredError       = "%s is required"
-
-	duplicateErrorCode = "23505"
+	commonFieldError        = "%s is not valid"
+	emailFormatError        = "Invalid email format"
+	emailDuplicateError     = "Email already exists"
+	minError                = "%s must be longer than %s"
+	passwordEqError         = "Passwords do not match"
```
olivierobert: After some more Go conventions reading, errors should always be in lowercase: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

helpers/user/current_user.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/gutakk/go-google-scraper/models"
+)
+
+func GetCurrentUser(c *gin.Context) models.User {
+	currentUser := c.MustGet("currentUser")
+
+	if currentUser != nil {
+		return currentUser.(models.User)
```
olivierobert: I am not sure why using [a type assertion](https://golang.org/ref/spec#Type_assertions) here as per the usage in the controller:

```go
currentUser := helpers.GetCurrentUser(c)
```

With the type assertion, shouldn't this be:

```go
currentUser, ok := helpers.GetCurrentUser(c)
```

controllers/keyword_test.go
```diff
@@ -0,0 +1,131 @@
...
+	engine := testConfig.GetRouter(true)
+	new(KeywordController).applyRoutes(EnsureAuthenticatedUserGroup(engine))
+
+	response := testHttp.PerformRequest(engine, "POST", "/keyword", nil, nil)
+
+	assert.Equal(t, http.StatusFound, response.Code)
+	assert.Equal(t, "/login", response.Header().Get("Location"))
+}
+
+func TestDisplayKeywordWithAuthenticatedUser(t *testing.T) {
```
olivierobert: Minor thing but shouldn't we group all tests cases for display and upload together. Now it seems messy which makes it hard to follow:

```
upload - TestUploadKeywordWithValidParams
upload - TestUploadKeywordWithBlankPayload
display - TestDisplayKeywordWithGuestUser
upload - TestUploadKeywordWithGuestUser
display - TestDisplayKeywordWithAuthenticatedUser
```

I am not sure now if `TestUploadKeywordWithValidParams` and `TestUploadKeywordWithBlankPayload` are for guest or authenticated users unless I check carefully the whole files.

It's more standard to group tests by:
- Order of appearance of the methods in the file under tests. Since the simply method appears first, then all tests for display should be first. Then the tests for upload can come after that.
- Happy path before sad path: so tests authenticated user should come before tests with no authenticated users within the group of tests for each method.

helpers/error_handler/error_message.go
gutakk: I also don't have better solution for this that's why I provided generic message ATM 😅

helpers/user/current_user.go
gutakk: Because `MustGet("currentUser")` return interface type so I need to cast it to `models.User` type so I can use its attribute. As `currentUser` can be any type so I added the type check in 891f46f42d4ef5ab6165819c5744526a5bd28996

helpers/error_handler/error_message.go
gutakk: Fixed in 43464969e9fc3212187adb73332eb81732a3e059

controllers/keyword_test.go
gutakk: Good catch Fixed in 423332f7a1b2b4fa1a7f52edd71809c496ddafaa.

Maybe should this conversation be a part of our Go convention?

controllers/keyword_test.go
gutakk: Good catch. Finally I found the way to save a cookie as a fixture in 4f3d586812302dec2af69154f873a02ea3475a22

controllers/keyword_test.go
olivierobert: This seems more like general best practice for all languages but if you feel it's needed to specify for Go then go for it.

helpers/user/current_user.go
olivierobert: Now with the type check, it does make more sense 😁

helpers/str/str.go
```diff
@@ -0,0 +1,13 @@
+package str
```
olivierobert: Instead of `str` `text` would be more descriptive 💭

helpers/html/render_html.go
```diff
@@ -4,13 +4,14 @@ import (
...
+	"github.com/gutakk/go-google-scraper/helpers/str"
 )
 
 func RenderWithError(c *gin.Context, status int, view string, title string, err error, data map[string]interface{}) {
 	ginview.HTML(c, status, view, gin.H{
-		"title":  title,
-		"errors": err.Error(),
-		"data":   data,
+		"title": title,
+		"error": str.CapitalizeFirst(err.Error()),
```
olivierobert: Not sure if `CapitalizeFirst` is really needed. The capitalization is more of a UI concern and could be solved with CSS 😄 Say, it was an API then it might make sense but for a web app, showing humane and well-formatted errors can be solved in the UI.

controllers/keyword_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	response := testHttp.PerformRequest(engine, "GET", "/keyword", nil, nil)
+
+	assert.Equal(t, http.StatusFound, response.Code)
+	assert.Equal(t, "/login", response.Header().Get("Location"))
+}
+
+func (s *KeywordDbTestSuite) TestUploadKeywordWithAuthenticatedUserAndValidParams() {
+	headers, payload := testFile.CreateMultipartPayload("tests/fixture/adword_keywords.csv")
+	cookie := fixture.GetCookie("user_id", s.userID)
```
olivierobert: I am not sure if it should be called `GetCookie` as it seems to actually generate encoded cookie data unless I am missing something in the helper code 🤔 So `fixture.GenerateCookie("user_id", s.userID)` might make more sense.

tests/csv/adword_keywords.csv
```diff
@@ -0,0 +1 @@
+AWS
```
olivierobert: This is a duplicate of the file `fixture/adword_keywords.csv` so it can be removed

olivierobert (commented):
Minor fixes to and then good to go 💪

controllers/keyword_test.go
gutakk: Make more sense. Fixed in c2abe467ca7f77b75aeb578ec9b69285bc0026ab

tests/csv/adword_keywords.csv
gutakk: 😮 My bad. Fixed in 4cc2fbcd1abd04500ccc952eec2ae4864c9c9bef

helpers/html/render_html.go
gutakk: Fixed in 2190a389763d3c2b8f161a17240f653665c92f63
I can't find any resources but I assume that Go unit test doesn't include css stuff. Thus I also need to update the unit test.

helpers/str/str.go
gutakk: According to this [comment](https://github.com/gutakk/go-google-scraper/pull/25#discussion_r536014123). I removed `str package`

helpers/html/render_html.go
olivierobert: In a unit test, the only test for lowercase, it's fine. For UI test, we also test for [case insensitivity](https://til.hashrocket.com/posts/85bda9479b-expect-a-case-insensitive-match-) to avoid tests breaking when the UI changes.

olivierobert (approved):

PR #27, opened 2020-12-03
As a user, I can view my uploaded keywords

### LINKED ISSUE #26: As a user, I can see uploaded keywords
### Acceptance Criteria
- Implement UI to display uploaded keywords
- Include unit test

gutakk:
## What happened
- [x] Add Keyword list table
- [x] Add nav skeleton

## Insight
- [x] Add keyword service to separate logic from model
- [x] As `gorm` provide both error string and pgconn error. Thus add the check condition for database error message

## Proof Of Work
- Keyword list table
Screen Shot 2563-12-04 at 18 04 50


assets/stylesheets/base/_form.scss
```diff
@@ -1,3 +1,7 @@
+form {
+  @apply mt-8;
```
olivierobert, line 2: Is this for all screens size? Have you checked on small screens?

assets/stylesheets/components/_nav.scss
```diff
@@ -0,0 +1,3 @@
+nav {
```
olivierobert: Components must always have a class name. Styles base tags are in `stylesheets/base`

assets/stylesheets/screens/keyword.scss
```diff
@@ -0,0 +1,4 @@
+.keyword-upload-form {
```
olivierobert: Prefer naming it `.form-keyword-upload` as per [Compass](https://nimblehq.co/compass/development/code-conventions/css/):

image

templates/layouts/application.html
```diff
@@ -9,11 +9,16 @@
   </head>
 
   <body>
-    <main class="app-main">
-      <div class="app-main__content">
-        {{ template "content" . }}
-      </div>
-    </main>
+    <div class="app">
+      {{ include "partials/navbar" }}
```
olivierobert: Is there really a need for this div at all? Would just staying the body tag be sufficient?

Avoid having a div wrapping all components. Instead, prefer having a stack of components:

```
<body>
  {{ include "partials/navbar" }}

  <main class="app-main">
  </main>

  {{ include "partials/footer" }}
</body>
```

This layout is more flexible as it allows to fit pretty much any design where the header and footer are of different sizes or positioning.

templates/keyword.html
```diff
@@ -1,15 +1,48 @@
...
+    <button type="submit">
+      Upload
+    </button>
+  </form>
 
-<h2>Upload CSV</h2>
+  {{ if .data.keywords }}
+    <div class="mt-8 -my-2 overflow-x-auto sm:-mx-6 lg:-mx-8">
+      <div class="py-2 align-middle inline-block min-w-full sm:px-6 lg:px-8">
+        <div class="shadow overflow-hidden border-b border-gray-200 sm:rounded-lg">
```
olivierobert: That seems like a lot of nested divs 😵 You have a case of divitis 😷 That issue is often caused by wanting to use all CSS helpers. It's also okay to mix helpers with custom styles:

image

Also prefer creating a custom component to wrap the Tailwin styles:

```
<div class="list-keyword">
  <table class="list-keyword__table"></table>
</div>
```

helpers/error_handler/error_message.go
```diff
@@ -43,13 +43,17 @@ func ValidationErrorMessage(err FieldError) error {
 
 // TODO: Improve later as this feel brittle
 func DatabaseErrorMessage(err error) error {
-	pgErr := err.(*pgconn.PgError)
-
-	switch pgErr.Code {
-	case foreignKeyErrorCode:
+	pgErr, isPgErr := err.(*pgconn.PgError)
+
+	if isPgErr {
```
olivierobert, line 48: What are the other cases of errors?

models/keyword.go
```diff
@@ -1,92 +1,37 @@
...
 	UserID  uint
 	User    User
 }
 
-func ReadFile(filename string) ([]string, error) {
-	csvfile, openErr := os.Open(filename)
-	if openErr != nil {
-		return nil, errors.New(somethingWentWrongError)
-	}
+func GetKeywords(condition map[string]interface{}) ([]Keyword, error) {
```
olivierobert: Since there is a condition, shouldn't it be `GetKeywordsBy` to be consistent with `FindUserBy`?

services/keyword_service/keyword.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	fileFormatError         = "file must be CSV format"
+	fileLengthError         = "CSV file must contain between 1 to 1000 keywords"
+	invalidDataError        = "invalid data"
+	somethingWentWrongError = "something went wrong, please try again"
+)
+
+type Keyword struct {
```
olivierobert: Should this struct be called `Keyword` as it's a bit confusing with the `Keyword` model struct 🤔 This struct is for KeywordService so it could be named after it?

olivierobert (changes requested):
More frontend-oriented issues this time 💅

helpers/error_handler/error_message.go
gutakk: As I read `gorm` source code, `gorm` provides only 2 error types which is `*errors.errorString` by gorm itself and `*pgconn.PgError` for other errors that gorm doesn't handle. That's why I use just only if else in this case 😅

models/keyword.go
gutakk: Good catch. Fixed in 03e19f9187f1c95d2a64d46b5b9c727cc46eda49

services/keyword_service/keyword.go
gutakk: I think it is confusing too but I cannot think of the good name. Thanks for pointing this out
Fixed in e60715e61314131376aa6ffc43039aa45388884d

assets/stylesheets/components/_nav.scss
gutakk: Fixed in 32daba55f936ed3126e46aaf0f36c8969c34354a

assets/stylesheets/base/_form.scss
gutakk: Yes this is for all screens size. I checked with small screens such as `800x600` `1024x768` `1280x720` `1366x768` or mobile device, the margin looks fine.

assets/stylesheets/screens/keyword.scss
gutakk: My bad 😭
Fixed in 90103ade0c4bc32c947c6801582409aede6c5317

templates/layouts/application.html
gutakk: Because I need `flex display` for navbar and app but not for auth.
if wrap everything in body tag and `display flex` in the body, this will effect the auth layout also.

What do you think?

templates/keyword.html
gutakk: Thank you for clarify this
Fixed in 85fe1a1de436bd5c897893b097b042e8359e22ac

templates/layouts/application.html
olivierobert: Actually, I understand the issue now and it also relates to another issue with screen stylesheets.

In the layout HTML, the body tag just have dynamic class names based on the controller/route name:

```
<body class="register show">
<body class="keyword show">
```

So that it can be used in screen stylesheets:

```
body.keyword .form-keyword-upload {

}

body.keyword .list-keyword {

}
```

This way the styles are only scoped to each page/screen.

In rails, that's what we use in the layout ERB/SLIM files:

```
body class="#{body_class_name}"
```

Which maps to this helper:

https://github.com/nimblehq/mia-web-backup/blob/64c275874fa085455ebb3bb4d1f7c09bf80da49f/app/helpers/css_helper.rb#L1-L11

So you need to find a similar way for Gin

assets/stylesheets/screens/keyword.scss
```diff
@@ -0,0 +1,12 @@
+.form-keyword-upload {
```
olivierobert: These styles are not scoped to a specific screen. See my below comment on `layout/application.html` to resolve this.

controllers/keyword.go
```diff
@@ -32,44 +32,65 @@ func (k *KeywordController) applyRoutes(engine *gin.RouterGroup) {
...
+	if validateTypeErr != nil {
+		html.RenderWithError(c, http.StatusBadRequest, keywordView, keywordTitle, validateTypeErr, data)
 		return
 	}
 
-	filename := models.UploadFile(c, form.File)
-	record, readFileErr := models.ReadFile(filename)
+	filename := keywordService.UploadFile(c, form.File)
+
+	record, readFileErr := keywordService.ReadFile(filename)
```
olivierobert: Naming this `record` is a bit confusing now with hindsight as it's a slice of strings 🤔  Even in the method `ReadFile`, using `record` is misleading. I would not advise using the plural `records` as plurals, in general, should be avoided to avoid typos. How about `wordList` (not to confuse with `keywordList` since these are raw words arsed from teh CSV file) or `parsedKeywordList`?

olivierobert (changes requested):

controllers/keyword.go
gutakk: Agree with you. Fixed in 13b2e6b4c493d3dc0f35d548b634c49daa4ebb74

templates/layouts/application.html
gutakk: Add controller class name to body in 543c36deb803b7d8024c6cd43fe795d26ed3c039
Update stylesheet in 6cdf6b77decf45ff5879d3fd7a8714a18b0b2c51

assets/stylesheets/screens/keyword.scss
gutakk: Fixed in this [comment](https://github.com/gutakk/go-google-scraper/pull/27#discussion_r539145928)

models/keyword.go
```diff
@@ -1,92 +1,37 @@
...
-	return nil
-}
-
-func ValidateFileType(fileType string) error {
-	if fileType != "text/csv" {
-		return errors.New(fileFormatError)
-	}
-	return nil
+func (k *Keyword) FormattedCreatedAt() string {
+	return k.CreatedAt.Format("December 2, 2006")
```
olivierobert: Wait, I did not see that before, why is there an actual date here 😳 Shouldn't be some pattern like `MM D, YYYY`?

models/keyword.go
```diff
@@ -1,92 +1,37 @@
...
-	}
-	return nil
-}
-
-func ValidateFileType(fileType string) error {
-	if fileType != "text/csv" {
-		return errors.New(fileFormatError)
-	}
-	return nil
+func (k *Keyword) FormattedCreatedAt() string {
```
olivierobert, line 35: This is a view concern so it would be better to have a presenter module instead of adding that to the model. This is an idea for further improvement.

models/keyword_test.go
```diff
@@ -69,65 +63,112 @@ func (s *KeywordDBTestSuite) TestSaveKeywordsWithValidParams() {
...
+	result, err := GetKeywordsBy(condition)
+
+	assert.Equal(s.T(), 1, len(result))
+	assert.Equal(s.T(), keyword.Keyword, result[0].Keyword)
+	assert.Equal(s.T(), nil, err)
 }
 
-func TestValidateCSVLengthWithGreaterThanMaxRowAllowed(t *testing.T) {
-	result := ValidateCSVLength(1001)
+func (s *KeywordDBTestSuite) TestGetKeywordsByWithFalsyCondition() {
```
olivierobert: I am not sure that the new method name `FalsyCondition`, `NilCondition`... help. I have to go check each test very much in details to understand the naming 🤔

models/keyword_test.go
```diff
@@ -69,65 +63,112 @@ func (s *KeywordDBTestSuite) TestSaveKeywordsWithValidParams() {
...
+	keyword := Keyword{UserID: s.userID, Keyword: faker.Name()}
+	db.GetDB().Create(&keyword)
 
-	assert.Equal(t, "CSV file must contain between 1 to 1000 keywords", result.Error())
+	condition := make(map[string]interface{})
+	condition["keyword"] = keyword.Keyword + "test"
```
olivierobert: To make this test easier, I would recommend to make it more obvious:

```go
keyword := Keyword{UserID: s.userID, Keyword: "Hazard"}
db.GetDB().Create(&keyword)

condition := make(map[string]interface{})
condition["keyword"] = "invalid"
```

`"invalid"` clearly documents that this value is wrong. This is similar as done 
below with `condition["unknown_column"]`.  `unknown_column` clearly conveys it's a wrong test data.

 Also to avoid naming conflict (in case `faker.Name()` generate "invalid" 🙃 ), it's better to hard code a keyword. By using the same fade data as above "Hazard", it also makes the tests easier to understand as a whole.

models/keyword.go
gutakk: Golang `Format` function can receive actual date and convert it into pattern for us.

For more information
https://programming.guide/go/format-parse-string-time-date-example.html
https://golang.org/pkg/time/#Time.Format

models/keyword.go
gutakk: Noted

models/keyword_test.go
gutakk: Could you please explain more about this?

models/keyword_test.go
gutakk: Understood. Fixed in 367212fd6ff880081defa56343a20731fc057bb0

models/keyword.go
olivierobert: Thanks for clarifying. Then I guess we should agree on which date to use 🤔  The examples shown seem to pick Jan 2nd, not Dec 2nd (not sure why this date by the way).

models/keyword_test.go
olivierobert: When reading the test method name `TestGetKeywordsByWithFalsyCondition`, I don't understand which case we are testing. So the name is not descriptive enough. 

That test is for retrieving the list of keywords when a keyword does not exist. So `TestGetKeywordsByWithInvalidKeyword` would be easier to read and understand. 

Same for `TestGetKeywordsByWithTruthyCondition` which could be replaced with `TestGetKeywordsByWithValidKeyword`. And `TestGetKeywordsByWithNilCondition` could be replaced with `TestGetKeywordsByWithBlankKeyword` or `TestGetKeywordsByWithoutKeyword`.

models/keyword_test.go
```diff
@@ -69,65 +63,112 @@ func (s *KeywordDBTestSuite) TestSaveKeywordsWithValidParams() {
...
+	result, err := SaveKeywords(bulkData)
 
 	assert.Equal(s.T(), nil, err)
 	assert.Equal(s.T(), 1, len(result))
 	assert.Equal(s.T(), "", result[0].Keyword)
 }
 
 func (s *KeywordDBTestSuite) TestSaveKeywordsWithEmptySlice() {
-	keywords := []string{}
+	bulkData := []Keyword{}
```
olivierobert: While plurals should be avoided, so is data 😆 `keywordList` would be better than `bulkData`.

gutakk: Fixed in 7f34c5bce857f061f6556f6256609313a5f02e05

models/keyword.go
gutakk: After some researched. I still not sure why only year `2006` and date `2` is the only values that can apply to the format and almost every example they use January.

So I will go with the format `January 2, 2006` in faf883ad5042b8143fb1e7af73a28cdd3b577ce1

models/keyword_test.go
gutakk: Thanks for clarifying. Fixed in 656fd51a651600d836bf427be928ffe27c4f83ad

olivierobert (approved):

PR #28, opened 2020-12-08
As a user, I can scrap data from google

### LINKED ISSUE #8: As a user, I can scrap data from google
### Acceptance Criteria
- Implement scraping feature
  - Backend to process scraping task
- Scraping result
  - Number of AdWords advertisers in the top position.
  - Total number of AdWords advertisers on the page.
  - URLs of the AdWords advertisers in the top position.
  - Number of the non-AdWords results on the page.
  - URLs of the non-AdWords results on the page.
  - Total number of links (all of them) on the page
  - HTML code of the page/cache of the page.
- Include unit test

gutakk:
## What happened
- Add google scraping as background job
- Add Redis
- Separate between test and development environment

## Insight
- [x] Add Redis for both test and development env
- [x] Add background job for google scraping (1 job per keyword) as a worker
- [x] Avoid mass-searching limitation by adding `User-Agent` and sleep each worker for 1 second after finish its job
- [x] Update run test command to run on test env
- [x] Trigger to enqueue google scraping job in keyword service after save keyword to database
- [x] Add scraping result column to keywords table
- [x] Add keyword status ENUM (`pending`, `processing`, `processed` and `failed`)
- [x] Add failed_reason column when job perform unsuccessfully
- [x] Max google scraping job fail is 3

## Proof Of Work
- Enqueue scraping job after upload csv file (app)
<img width="648" alt="Screen Shot 2563-12-15 at 15 51 04" src="https://user-images.githubusercontent.com/29707647/102192435-5f2eb580-3eed-11eb-8f2f-9e5342df7c7d.png">

- Process job in the background (worker)
<img width="641" alt="Screen Shot 2563-12-15 at 15 51 51" src="https://user-images.githubusercontent.com/29707647/102192514-7372b280-3eed-11eb-9710-7415d4d95ab6.png">

- Database after processed
```
// SELECT id, created_at, keyword, status, links_count, non_adwords_count, top_position_adwords_count, total_adwords_count, user_id FROM keywords;

 id |          created_at           |      keyword      |  status   | links_count | non_adwords_count | top_position_adwords_count | total_adwords_count | user_id 
----+-------------------------------+-------------------+-----------+-------------+-------------------+----------------------------+---------------------+---------
 17 | 2020-12-15 08:52:42.065812+00 | Eden Hazard       | processed |         117 |                 8 |                          0 |                   0 |       1
 18 | 2020-12-15 08:52:42.065812+00 | Lionel Messi      | processed |         112 |                 9 |                          0 |                   0 |       1
 19 | 2020-12-15 08:52:42.065812+00 | Cristiano Ronaldo | processed |         122 |                 9 |                          0 |                   0 |       1
 20 | 2020-12-15 08:52:42.065812+00 | Kylian Mbappe     | processed |         113 |                 9 |                          0 |                   0 |       1
 21 | 2020-12-15 08:52:42.065812+00 | Neymar            | processed |         115 |                 8 |                          0 |                   0 |       1
 22 | 2020-12-15 08:52:42.065812+00 | Kevin De Bruyne   | processed |         122 |                 9 |                          0 |                   0 |       1
(6 rows)
```

```
// select id, keyword, non_adword_links from keywords;

 id |      keyword      |                                                                                                                                                                                   
                                                                                                                                         non_adword_links                                                   
                                                                                                                                                                                                            
                                                              
----+-------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------
 17 | Eden Hazard       | ["https://th.wikipedia.org/wiki/%E0%B9%80%E0%B8%AD%E0%B9%81%E0%B8%94%E0%B8%99_%E0%B8%AD%E0%B8%B2%E0%B8%8B%E0%B8%B2%E0%B8%A3%E0%B9%8C", "https://en.wikipedia.org/wiki/Eden_Hazard"
, "https://www.transfermarkt.com/eden-hazard/profil/spieler/50202", "https://www.pinterest.com/mayurespu/eden-hazard/", "https://twitter.com/hazardeden10", "https://th-th.facebook.com/edenhazard/", "https
://www.instagram.com/hazardeden_10/?hl=th", "https://www.pinterest.dk/qazwxe9999/eden-hazard/"]
 18 | Lionel Messi      | ["https://th.wikipedia.org/wiki/%E0%B8%A5%E0%B8%B4%E0%B9%82%E0%B8%AD%E0%B9%80%E0%B8%99%E0%B8%A5_%E0%B9%80%E0%B8%A1%E0%B8%AA%E0%B8%8B%E0%B8%B4", "https://en.wikipedia.org/wiki/Lio
nel_Messi", "https://www.pinterest.com/queenbeerr/lionel-messi/", "https://www.transfermarkt.com/lionel-messi/profil/spieler/28003", "https://www.instagram.com/leomessi/", "https://www.forbes.com/profile/
lionel-messi/", "https://www.fcbarcelona.com/en/football/first-team/players/4974/lionel-messi", "https://th-th.facebook.com/leomessi", "https://www.facebook.com/leomessi"]
 19 | Cristiano Ronaldo | ["https://th.wikipedia.org/wiki/%E0%B8%84%E0%B8%A3%E0%B8%B4%E0%B8%AA%E0%B9%80%E0%B8%95%E0%B8%B5%E0%B8%A2%E0%B9%82%E0%B8%99_%E0%B9%82%E0%B8%A3%E0%B8%99%E0%B8%B1%E0%B8%A5%E0%B9%82%
E0%B8%94", "https://en.wikipedia.org/wiki/Cristiano_Ronaldo", "https://th-th.facebook.com/Cristiano", "https://www.facebook.com/Cristiano", "https://twitter.com/cristiano", "https://www.instagram.com/cris
tiano/", "https://www.cristianoronaldo.com/", "https://www.transfermarkt.com/cristiano-ronaldo/profil/spieler/8198", "https://www.pinterest.com/pin/397161260884459437/"]
 20 | Kylian Mbappe     | ["https://en.wikipedia.org/wiki/Kylian_Mbapp%C3%A9", "https://www.transfermarkt.com/kylian-mbappe/profil/spieler/342229", "https://www.instagram.com/k.mbappe/", "https://www.live
rpoolfc.com/news/media-watch/419506-liverpool-s-interest-in-kylian-mbappe-real", "https://www.youtube.com/watch?v=eFq8d4Fpsqo&list=PLZfhSmptyNbEyom2XTfkaPtjXU8h9dHu9&index=137", "https://en.psg.fr/teams/f
irst-team/squad/kylian-mbappe", "https://www.independent.co.uk/topic/kylian-mbappe", "https://www.dailymail.co.uk/sport/kylian-mbappe/index.html", "https://www.sportskeeda.com/player/kylian-mbappe"]
 21 | Neymar            | ["https://th.wikipedia.org/wiki/%E0%B9%80%E0%B8%99%E0%B8%A2%E0%B9%8C%E0%B8%A1%E0%B8%B2%E0%B8%A3%E0%B9%8C", "https://en.wikipedia.org/wiki/Neymar", "https://www.pinterest.com/sutt
icha864/neymar/", "https://www.transfermarkt.com/neymar/profil/spieler/68290", "https://www.facebook.com/neymarjr/", "https://shopee.co.th/search?keyword=%23neymar", "https://www.neymarjr.com/", "https://
edition.cnn.com/2020/12/14/football/neymar-injury-psg-lyon-spt-intl/index.html"]
 22 | Kevin De Bruyne   | ["https://th.wikipedia.org/wiki/%E0%B9%80%E0%B8%81%E0%B8%9F%E0%B8%B4%E0%B8%99_%E0%B9%80%E0%B8%94%E0%B8%AD_%E0%B9%80%E0%B8%9A%E0%B8%A3%E0%B8%A2%E0%B9%80%E0%B8%99%E0%B8%AD", "https
://en.wikipedia.org/wiki/Kevin_De_Bruyne", "https://www.transfermarkt.com/kevin-de-bruyne/profil/spieler/88755", "https://www.instagram.com/kevindebruyne/?hl=th", "https://www.premierleague.com/players/42
88/Kevin-De-Bruyne/overview", "https://www.premierleague.com/players/4288/Kevin-De-Bruyne/stats", "https://twitter.com/debruynekev", "https://www.whoscored.com/Players/73084/Show/Kevin-De-Bruyne", "https:
//www.manchestereveningnews.co.uk/all-about/kevin-de-bruyne"]
(6 rows)
```


services/google_scraping_service/parser.go
```diff
@@ -0,0 +1,75 @@
...
+		return ScrapingResult{}, err
+	}
+
+	scrapingResult := ScrapingResult{
+		LinksCount:             g.countLinks(doc),
+		NonAdwordsCount:        g.countNonAdwords(doc),
+		NonAdwordLinks:         g.fetchNonAdwordLinks(doc),
+		TopPostionAdwordsCount: g.countTopPositionAdwords(doc),
+		TopPositionAdwordLinks: g.fetchTopPositionAdwordLinks(doc),
+		TotalAdwordsCount:      g.countTotalAdwords(doc),
```
olivierobert: It might be good as a nice-to-have improvement to look into goroutines to parse this asynchronously, now it's very linear

workers/jobs/scraping.go
```diff
@@ -0,0 +1,20 @@
+package jobs
```
olivierobert: Avoid `-ing` in the naming job, prefer the imperative tone `scrap_keyword_job.go` or using a noun `keyword_scraper.go`

olivierobert (commented):
Some early feedback

gutakk:
@olivierobert I split the UI implementation to another [issue](https://github.com/gutakk/go-google-scraper/issues/29) as I think this PR will has a lot of changes by just only backend.

workers/jobs/scraping.go
gutakk: Fixed in 60390a66288acf8ac626261273fbf39b8aca1115

services/google_scraping_service/parser.go
gutakk: Improved in 89f24867d743bdfa366afba6ae2f85715033ad50

gutakk: Found an issue when working with background job. Some result not parsing properly, so I will go with the linear style for now and improve it at the end or maybe in the separated branch.

gutakk: > Found an issue when working with background job. Some result not parsing properly, so I will go with the linear style for now and improve it at the end or maybe in the separated branch.

Fixed this issue. It is not about goroutine stuff. The issue caused from my inconsistent css selector, so now I went back to goroutine way.

controllers/keyword.go
```diff
@@ -72,12 +73,18 @@ func (k *KeywordController) uploadKeyword(c *gin.Context) {
...
 
 	// Save keywords to database
-	_, saveKeywordsErr := keywordService.Save(parsedKeywordList)
+	savedKeywords, saveKeywordsErr := keywordService.Save(parsedKeywordList)
 	if saveKeywordsErr != nil {
 		html.RenderWithError(c, http.StatusBadRequest, keywordView, keywordTitle, saveKeywordsErr, data)
 		return
 	}
 
+	enqueueErr := google_scraping_service.EnqueueScrapingJob(savedKeywords)
```
olivierobert: Prefer enqueuing jobs in the service module. Controllers should not manage complex business logic. Saving keywords and triggering the scraping work is hard business logic that belongs in the service. Think that if there is a CLI app, it can use the same service and it would behave the same because the business logic is not in the controller.

migration/migration.go
```diff
@@ -15,9 +15,21 @@ func Migrate(db *gorm.DB) {
...
+	InitKeywordStatusEnum(db)
+
 	if err := db.AutoMigrate(&models.Keyword{}); err != nil {
 		log.Fatal(fmt.Sprintf("Failed to migrate %v", err))
 	} else {
 		log.Print("Migrate keyword schema successfully")
 	}
 }
+
+func InitKeywordStatusEnum(db *gorm.DB) {
```
olivierobert, line 28: Are all migrations in the same files? Can't they be split into several files like in Rails?

controllers/keyword.go
gutakk: I got your point. Fixed in 9bc1bb821b0063ab483b6d84f88afc9f509ff04d

migration/migration.go
gutakk: I follow `GORM` tutorial here. I think we can improve this but I think it would be better to done this in another PR.
What do you think?

gutakk: Yes we can but I think we should refactor this in another PR because this PR is way too big.
What do you think?

olivierobert: Of course, always move all refactors to chore branches. Easier to review.

controllers/keyword.go
```diff
@@ -72,7 +72,7 @@ func (k *KeywordController) uploadKeyword(c *gin.Context) {
 	}
 
 	// Save keywords to database
-	_, saveKeywordsErr := keywordService.Save(parsedKeywordList)
+	saveKeywordsErr := keywordService.SaveAndScrape(parsedKeywordList)
```
olivierobert: Not sure about the need to add `Scrape` to the method name. What if it needs to send an email too, would it be `SaveScrapeAndNotify`. I find `keywordService.Save` enough as it moves the complexity to the service module and it allows more flexibility.

db/redis.go
```diff
@@ -0,0 +1,42 @@
...
+
+var RedisPool *redis.Pool
+
+func GenerateRedisPool(address string) {
+	pool := &redis.Pool{
+		MaxActive: 5,
+		MaxIdle:   5,
+		Wait:      true,
+		Dial: func() (redis.Conn, error) {
+			return redis.Dial("tcp", address)
```
olivierobert: Why not using env variables making it a configuration concern (thus removing the need to send an argument):

```go
host := os.Getenv("REDIS_TEST_HOST"
port := os.Getenv("REDIS_TEST_PORT")

return redis.Dial("tcp", host + ":" + port)
```

Now `db.GenerateRedisPool("localhost:6380")` is duplicated.

models/keyword.go
```diff
@@ -32,6 +64,15 @@ func SaveKeywords(keywords []Keyword) ([]Keyword, error) {
 	return keywords, nil
 }
 
+func UpdateKeywordByID(keywordID uint, newKeyword Keyword) error {
```
olivierobert: Minor naming preference to `UpdateKeyword` as it's implicit an identifier would be needed to update a keyword. Unless there is another method to update a keyword, I would recommend for simplicity.

models/keyword_test.go
```diff
@@ -62,6 +64,46 @@ func (s *KeywordDBTestSuite) TestSaveKeywordsWithValidParams() {
 	assert.Equal(s.T(), "Mbappe", result[4].Keyword)
 }
 
+func (s *KeywordDBTestSuite) TestSaveKeywordsWithSingleValidParams() {
```
olivierobert: The need to have both single keyword and multiple keyword tests is a code smell IMO as it causes lots of complexity.

It would be simpler to have a method to save **one** keyword:

```go
func SaveKeywords(keyword Keyword) (Keyword, error) {}
```

So in the `keyword_service` package, instead of building a slice of keyword in `func (k *KeywordService) SaveAndScrape(parsedKeywordList []string) error`, it would better to loop over the slide of strings `parsedKeywordList` and save one keyword at a time. A nice-to-have here (but must-have in a production-ready app) would be to wrap all the different database creation calls into [a DB transaction](https://dev.to/techschoolguru/a-clean-way-to-implement-database-transaction-in-golang-2ba).

Even the scraping enqueuing should be done at an atomic level - at each keyword level. Schedule one keyword job is easier than scheduling multiple at once.

services/google_scraping_service/db_updater.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func UpdateKeywordWithParsingResult(keywordID uint, parsingResult ParsingResult) error {
+	nonAdwordLinks, nonAdwordLinksParsingErr := json.Marshal(parsingResult.NonAdwordLinks)
+	if nonAdwordLinksParsingErr != nil {
```
olivierobert: Why is it necessary to use `json.Marshal`. If there is no data, couldn't this be just an empty slice `[]string`?

services/google_scraping_service/job_enqueuer_test.go
```diff
@@ -0,0 +1,77 @@
...
+	conn := db.GetRedisPool().Get()
+	defer conn.Close()
+
+	redisKey := testDB.RedisKeyJobs("go-google-scraper", "scraping")
+	var jobArgList []string
+
+	for i := 0; i < len(savedKeywordList); i++ {
+		rawJSON, errs := redis.Bytes(conn.Do("RPOP", redisKey))
+		if errs != nil {
+			panic("could RPOP from job queue: " + errs.Error())
```
olivierobert: What is this error case about?

services/google_scraping_service/parser.go
```diff
@@ -0,0 +1,97 @@
...
+	HtmlCode               string
+	LinksCount             int
+	NonAdwordsCount        int
+	NonAdwordLinks         []string
+	TopPostionAdwordsCount int
+	TopPositionAdwordLinks []string
+	TotalAdwordsCount      int
+}
+
+var ParseGoogleResponse = func(googleResp *http.Response) (ParsingResult, error) {
```
olivierobert: Why setting the function to a variable?

workers/jobs/google_scraper.go
```diff
@@ -0,0 +1,92 @@
...
+	log.Printf("Job %v for keyword %v done in %v", jobName, keyword, end.String())
+
+	time.Sleep(1 * time.Second)
+	return nil
+}
+
+// Update status to failed when (jobFails + 1) reach MaxFails. Note: Job won't retry if jobFails reach MaxFails
+// So this need to be done at jobFails + 1
+func updateStatusToFailed(jobFails int64, jobName string, keywordID uint, keyword string, err error) {
+	if int(jobFails+1) >= MaxFails {
```
olivierobert: Minor thing but shouldn't this be `(int(jobFails) + 1) >= MaxFails? There is no need to cast ` inside int().

workers/jobs/google_scraper.go
```diff
@@ -0,0 +1,92 @@
...
+	"log"
+	"time"
+
+	"github.com/gocraft/work"
+	"github.com/gutakk/go-google-scraper/models"
+	"github.com/gutakk/go-google-scraper/services/google_scraping_service"
+)
+
+const (
+	MaxFails = 3
```
olivierobert, line 16: How about making this a configuration environment variable? See the [12 Factor app](https://12factor.net/config) for reference.

workers/main.go
```diff
@@ -0,0 +1,44 @@
...
+var redisPool = &redis.Pool{
+	MaxActive: 5,
+	MaxIdle:   5,
+	Wait:      true,
+	Dial: func() (redis.Conn, error) {
+		return redis.Dial("tcp", "localhost:6379")
+	},
+}
+
+func main() {
```
olivierobert, line 31: Would this restart all pending jobs (not started and below `jobs.maxFails`) when the worker process is stopped and restarted?

migration/migration.go
gutakk: Added TODO in ac1536ae7cc35774de51b9ffe601c57cd62535f5

controllers/keyword.go
gutakk: Agree. Fixed in 72e166e732927c1d56f6e6f6ff698e2dd23dd763

db/redis.go
gutakk: Fixed in 8dddb821312aa28190dac06d1966b94832643d76

models/keyword.go
gutakk: Good catch. Fixed in 8b0e0a53487ce7c0d435103dff88489f4da6156b

models/keyword_test.go
gutakk: Could you please explain more why save keyword one by one in the transaction is better than save the whole keyword list without transaction? 🙏

services/google_scraping_service/db_updater.go
gutakk: Because `NonAdwordLinks` and `TopPositionAdwordLinks` type are `datatypes.JSON` from `gorm.io/datatypes` and the datatypes.JSON actual type is `[]byte`

services/google_scraping_service/job_enqueuer_test.go
gutakk: This error occur when you try to RPOP from empty queue in Redis.

db/redis.go
gutakk: The workflow failed remind me about why we cannot use environment variables in the test. Because we ignore all the .env file except .env.example so in the CI all the tests that need env variable will fail.

Is it acceptable to push `.env.test` to git repo?

olivierobert: Yes, it's okay to commit the file. Only the production one should not be committed.

services/google_scraping_service/parser.go
gutakk: Because I want to mock that function in the test.
There are another way to mock the function by using an interface or send the function as a parameter but I think we cannot do that in this case because this function is called by the worker instead of directly method call.

services/google_scraping_service/job_enqueuer_test.go
gutakk: Update error message in 3d1f90bf52e5d5cc7c75a396d865186032849458

db/redis.go
gutakk: Add `.env.debug` and `.env.test` in 26621e16d3203e0772e4b832c5da17b50236e1ce

workers/jobs/google_scraper.go
gutakk: Fixed in 6199b7fa3fb54e1d2c3115f2578e3de6c1e74ba1

workers/jobs/google_scraper.go
gutakk: In this case our production app is deployed to Heroku. What is the way should we do about env variables on production?
1. Add ENV directly to Heroku config
2. Create `.env` and `.env.release` in local machine and let Dockerfile copy that 2 files when build

gutakk:
@olivierobert
I forget to add worker Dockerfile for production.
Added in 9b9784414648671112e6a307d40214778fb980d1

workers/main.go
gutakk: Yes. here are the steps as following GIF
1. Start app
2. Uploads keywords and then app will enqueue all the job
3. Start worker
4. Worker executes pending jobs

lrUEs3Y3WJ

Makefile
```diff
@@ -10,5 +10,8 @@ env-setup:
 start-dev: env-setup
 	forego start
 
-test: env-setup
-	go test -v -p 1 ./...
+test:
+	docker-compose -f docker-compose.test.yml up -d
+	go test -v -p 1 -count=1 ./...
+	docker-compose -f docker-compose.test.yml down
+
```
andyduong1920: Small thing, should we delete the empty line here 🤔

main.go
```diff
@@ -12,8 +12,10 @@ import (
 
 func main() {
 	config.LoadEnv()
-	db := db.ConnectDB()
-	migration.Migrate(db)
+	database := db.ConnectDB()
+	migration.Migrate(database)
+
+	db.GenerateRedisPool()
```
andyduong1920: Just small thing :D for me, this is not `generate` something, it looks like `setup` the Redis pool, or `init` the Redis pool. What do you think @gutakk

db/redis.go
olivierobert: Think of updating the README files as well 

image

services/google_scraping_service/db_updater.go
olivierobert: Alright. Thanks for the explanation. I did miss the JSON data type for these columns 🙈

services/google_scraping_service/parser.go
olivierobert: When you explain about the mock, you refer to the use of Govcr?

workers/main.go
olivierobert: Then this solution seems very complete 🤗

models/keyword_test.go
olivierobert: Mostly because it's less complex and more flexible.

Think of a keyword as a resource with its own lifecycle. Each keyword must be first saved to the database and then a job must be enqueued. Each step must be happening for every new keyword. 

Currently, both steps are not done the same place. That's introducing complexity. There are two loops:
- One loop in `KeywordService` to create a list of a keyword to be saved.
- Second loop in `EnqueueScrapingJob` to enqueue each keyword job.

What I am recommending is to have a single loop in `KeywordService` to:
- Save each keyword
- Enqueue a new job for each keyword right after saving the keyword

services/google_scraping_service/job_enqueuer_test.go
```diff
@@ -0,0 +1,77 @@
+package google_scraping_service
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/gocraft/work"
+	"github.com/gomodule/redigo/redis"
+	"github.com/gutakk/go-google-scraper/db"
+	"github.com/gutakk/go-google-scraper/models"
```
andyduong1920: Is this one the internal package @gutakk `"github.com/gutakk/go-google-scraper/models"` Am I wrong?

db/redis.go
gutakk: Updated in 47a30a7178417b517806e5610cbf85cedca4b95b

services/google_scraping_service/parser.go
gutakk: For `parser` I want to mock it for jobs testing because I want to make sure that if parse fail and reach max fails job need to update keyword status to `failed`

For `request` also the same with parser and one more thing is I am able to use VCR when I test the valid job.

You can check this out 
https://github.com/gutakk/go-google-scraper/blob/47a30a7178417b517806e5610cbf85cedca4b95b/workers/jobs/google_scraper_test.go

workers/jobs/google_scraper.go
olivierobert: The first option would be the Heroku way 👍 Nothing wrong with the second option but it's good to use nice features like env variables management provided by Heroku.

Also just to clarify:

```
Create .env and .env.release in local machine
```

`local machine` would be the CI server 🧐

services/google_scraping_service/parser.go
```diff
@@ -0,0 +1,97 @@
...
+
+	go countBottomPositionAdwords(doc, countBottomPositionAdwordsCh)
+	go countLinks(doc, countLinksCh)
+	go countNonAdwords(doc, countNonAdwordsCh)
+	go countTopPositionAdwords(doc, countTopPositionAdwordsCh)
+
+	go fetchNonAdwordLinks(doc, fetchNonAdwordLinksCh)
+	go fetchTopPositionAdwordLinks(doc, fetchTopPositionAdwordLinksCh)
+
+	bottomPositionAdwordsCount := <-countBottomPositionAdwordsCh
```
andyduong1920: I think we could improve the performance at this place by using [select](https://tour.golang.org/concurrency/5) so we can have the concurrency process here, with the current implementation, the result comes in sequential, it needs to wait for `<- countBottomPositionAdwordsCh` has the value before going to `<- countTopPositionAdwordsCh`, same for other chans bellow

models/keyword_test.go
gutakk: - Updated SaveKeyword in model in ca9fa05f6358211522b467fcc0d5837d277a6090
- Updated service (without transaction) in e43d3a88a0b761c402ddd9d9e54d3fe4013fe562

gutakk: Added transaction in ee4c0a05658f1af67d3b74d0279325966de88f8d.
It looks weird but this is the only way I figured out.
@olivierobert Do you have any comment? or can we leave this PR without transaction?

services/google_scraping_service/job_enqueuer_test.go
gutakk: Yes it is. My module name is `github.com/gutakk/go-google-scraper` so any packages that has this prefix are internal.
I decide to refactor all of my code to follow our Golang convention in another PR, I already open an [issue](https://github.com/gutakk/go-google-scraper/issues/33)

But for the new file I will follow the convention right away.

migration/migration.go
olivierobert: Also, add a new chore task for this change too

models/keyword_test.go
olivierobert: It's kind of weird indeed as it requires to pass the transaction to the method `SaveKeyword` as an argument:

```go
models.SaveKeyword(keyword, tx)
```

I checked the docs and it seems to be the way to go but there might be better ways to do it. Up to you to keep it or not.

models/keyword.go
```diff
@@ -22,14 +54,29 @@ func GetKeywordsBy(condition map[string]interface{}) ([]Keyword, error) {
 	return keywords, nil
 }
 
-func SaveKeywords(keywords []Keyword) ([]Keyword, error) {
-	// Insert bulk data
-	result := db.GetDB().Create(&keywords)
+func SaveKeyword(keyword Keyword, tx ...*gorm.DB) (Keyword, error) {
+	var cnx *gorm.DB
+	if tx != nil {
+		cnx = tx[0]
```
olivierobert: That's the strangest part of the transaction implementation as I am not sure why we need to select the first item in the tx slide argument 🤔

models/keyword.go
```diff
@@ -22,14 +54,29 @@ func GetKeywordsBy(condition map[string]interface{}) ([]Keyword, error) {
...
+	} else {
+		cnx = db.GetDB()
+	}
+
+	result := cnx.Create(&keyword)
 	if result.Error != nil {
-		return nil, result.Error
+		return Keyword{}, result.Error
```
olivierobert, line 69: Why not returning nil as before?

models/keyword_test.go
```diff
@@ -43,59 +45,75 @@ func TestKeywordDBTestSuite(t *testing.T) {
...
+	assert.Equal(s.T(), []string{"test-non-ads-link"}, nonAdwordLinksVal)
+	assert.Equal(s.T(), 5, result.TopPositionAdwordsCount)
+	assert.Equal(s.T(), []string{"test-top-ads-link"}, topPositionAdwordLinksVal)
+	assert.Equal(s.T(), 25, result.TotalAdwordsCount)
+	assert.Equal(s.T(), "test-html", result.HtmlCode)
 }
 
-func (s *KeywordDBTestSuite) TestSaveKeywordsWithEmptySlice() {
-	keywordList := []Keyword{}
+func (s *KeywordDBTestSuite) TestSaveKeywordsWithKeywordValue() {
```
olivierobert: Shouldn't this be:

```suggestion
func (s *KeywordDBTestSuite) TestSaveKeywordsWithInvalidKeywordValue() {
```

services/google_scraping_service/db_updater.go
```diff
@@ -0,0 +1,53 @@
+package google_scraping_service
```
olivierobert: As discussed before, I recommend to renamed this service to `google_search_service`. It sounds better (NO `-ing`) and it's more flexible if new Search services are added.

services/google_scraping_service/job_enqueuer.go
```diff
@@ -0,0 +1,38 @@
...
+
+var EnqueueScrapingJob = func(savedKeyword models.Keyword) error {
+	if len(savedKeyword.Keyword) == 0 {
+		return errors.New(invalidKeyword)
+	}
+
+	enqueuer := work.NewEnqueuer("go-google-scraper", db.GetRedisPool())
+
+	job, err := enqueuer.Enqueue(
+		"scraping",
```
olivierobert: Again as an improvement to remove all `-ing`, how about renaming to "search"? It would map nicely with the new name of `google_search_service`.

services/google_scraping_service/job_enqueuer.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/gocraft/work"
+	"github.com/gutakk/go-google-scraper/db"
+	"github.com/gutakk/go-google-scraper/models"
+)
+
+const (
+	invalidKeyword = "invalid keyword"
+)
+
+var EnqueueScrapingJob = func(savedKeyword models.Keyword) error {
```
olivierobert: Related to the new name I am recommending, this method name could be changed to:

```suggestion
var EnqueueSearchJob = func(savedKeyword models.Keyword) error {
```

services/google_scraping_service/parser.go
olivierobert: I also would be interested to check if that changes anything here 😁

services/keyword_service/keyword_test.go
```diff
@@ -69,27 +88,41 @@ func (s *KeywordServiceDbTestSuite) TestGetAllWithInvalidUser() {
...
 	assert.Equal(s.T(), "invalid data", err.Error())
-	assert.Equal(s.T(), nil, result)
+}
+
+func (s *KeywordServiceDbTestSuite) TestSaveWithEnqueueJobError() {
+	enqueueScrapingJobFunc := google_scraping_service.EnqueueScrapingJob
+	google_scraping_service.EnqueueScrapingJob = func(savedKeyword models.Keyword) error {
+		return errors.New("mock enqueue scraping job error")
+	}
+	defer func() { google_scraping_service.EnqueueScrapingJob = enqueueScrapingJobFunc }()
```
olivierobert: SO that's the whole mocking setup, right?

workers/jobs/google_scraper_test.go
```diff
@@ -0,0 +1,207 @@
...
+		req.Header.Set("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36")
+
+		resp, err := client.Do(req)
+		if err != nil {
+			return nil, err
+		}
+
+		return resp, err
+	}
+	defer func() { google_scraping_service.Request = requestFunc }()
```
olivierobert: I don't get how this mock works as it seems to reimplement the whole Request method 🤔

olivierobert (commented):
Looks complete. I am recommending some improvements on the naming and I am not sure about the mocking setup at the moment.

migration/migration.go
gutakk: Issue added
https://github.com/gutakk/go-google-scraper/issues/34

models/keyword_test.go
gutakk: > It's kind of weird indeed as it requires to pass the transaction to the method `SaveKeyword` as an argument:
> 
> ```go
> models.SaveKeyword(keyword, tx)
> ```
> 
> I checked the docs and it seems to be the way to go but there might be better ways to do it. Up to you to keep it or not.

I think I would keep it as I think this kind of thing should has a transaction.

models/keyword.go
gutakk: Function that receive parameter like this `(tx ...*gorm.DB) called Variadic function and that allow you to pass 0 or more parameter to the function.
Here is the simple example
https://gobyexample.com/variadic-functions

models/keyword.go
gutakk: Because struct type does not allow to return nil
```go
cannot convert nil (untyped nil value) to Keyword
```

models/keyword_test.go
gutakk: My bad. Fixed in 5c34fff8dc68b0f26cda6725b34aa03f4a711292

services/google_scraping_service/db_updater.go
gutakk: Fixed in c31a4280dc801ab6a3c11aa295e54638900a9d8e

services/google_scraping_service/job_enqueuer.go
gutakk: Fixed in de9d39e0563cefb62d222ae4e9bcd9e4ea35aeb6

services/google_scraping_service/job_enqueuer.go
gutakk: Fixed in 5ec85f0f67f6e6f0b2e9d7839120284ac2b90be0

services/keyword_service/keyword_test.go
gutakk: Yes it is. The idea is override the function for testing purpose.

workers/jobs/google_scraper_test.go
gutakk: In this case, in order to use VCR I need to set `Transport` in the client and the `Request` function is called in `PerformSearchJob` and it is executed by worker so I cannot find the way how can I pass value to set the `Transport` in this case. Thus I need to override the request function to set the `Transport`.

gutakk: But I think as this is a unit test. I can just mock to return valid response (not to be response from google) and just check that job err is not nil as the function return `error`

This kind of test is more likely to be integration testing

What do you think?

Makefile
gutakk: Fixed in ef7f1923a720bc01ea943c0cba3da4cc5536b4d1

main.go
gutakk: Good catch. Fixed in c5a7ff194669df19d656fe7a67587e44b36f3094

services/google_scraping_service/job_enqueuer_test.go
gutakk: Updated import order in ffd39290782311b9da1a359d15b55420fdb30bf2

workers/jobs/google_scraper_test.go
gutakk: Fixed in def6a406845fa78cb7ca74d506b105cb3ffaad8d

services/google_scraping_service/parser.go
gutakk: Tried to use `select` in 297b97c8c20002c290d3c2ac810769c9b841f609

models/keyword.go
olivierobert: Oh yes, indeed 😓

workers/jobs/google_scraper_test.go
olivierobert: Indeed, just mocking the return value would be enough. That's what a mock should do in a unit test. I checked the fix but I still don't fully understand how it actually works 😓 

In `SetupTest`, `requestFunc` and `parsingFunc` are set to teh actua; implementation for `google_search_service.Request` and `google_search_service.ParseGoogleResponse`. So no mock implementation here, right?

https://github.com/gutakk/go-google-scraper/blob/ff0ee1455035e6a5649032ae7656306ab58efd1c/workers/jobs/google_scraper_test.go#L55-L59

And then the mocks are implemented in `TestPerformSearchJobWithValidJob`:

https://github.com/gutakk/go-google-scraper/blob/ff0ee1455035e6a5649032ae7656306ab58efd1c/workers/jobs/google_scraper_test.go#L80-L86

So the part I am not sure is how and why using `defer` 🤔

olivierobert: Actually, after some googling around, `defer` is used to restore back to the original/application implementation. I understand that `defer` part now. 

But I also understand now teh following is used to store the current implementation:

```go
s.requestFunc = google_search_service.Request 
s.parsingFunc = google_search_service.ParseGoogleResponse 
```

But only `TestPerformSearchJobWithValidJob` implements the mocks while the other tests like `TestPerformSearchJobWithoutKeywordID` do not or implement only one 🤔 It is kind of unexpected and could lead to hard-to-see and hard-to-solve issues. To be truly a unit test (so not an integration test), all test should implement the mocks/overrides the default implementation. Or the setup method should implement a general-purpose mock.

olivierobert: Also, as a recommendation, it would make `SetupTest` more readable by  externalizing to methods"

```go
func (s *KeywordScraperDBTestSuite) SetupTest() {
  // these methods could be shared to a test helpers module
  database, _ := testHelpers.SetiupDatabase()
  s.userID = testHelpers.createUser()
  
  // This can be a private method in this test file/package
  setupMocks()

  s.enqueuer = work.NewEnqueuer("test-job", db.GetRedisPool())
}
```

It would make tests easier to manage and maintain.

gutakk: Added setupMocks() in aeca2ef754d0dc4cbee5ae1ece1b3837bd432b05.
For another setup I will do it in the [test refactor issue](https://github.com/gutakk/go-google-scraper/issues/36).

models/keyword_test.go
olivierobert: I don't have a better option as well for now 😔 So it can be marked as a FIXME

models/keyword.go
olivierobert: I understood the argument type, but why I am not sure why we always pick the first item in the tx param 🙈 There is no loop over `tx` so I don't get why we need to use a variadic paam.

olivierobert (approved):
Looks good to me now. Thanks for the hard work 💪 

I am sure there are still some areas to improve on but it's good to merge and move forward.

models/keyword_test.go
gutakk: Added TODO in 1807813c580b9a0df5262c564930e63b3c9b8bd2

models/keyword.go
gutakk: In the first place, I thought I want transaction to be optional that's why I use variadic param but that's not make sense so Fixed in 1807813c580b9a0df5262c564930e63b3c9b8bd2

models/keyword_test.go
```diff
@@ -172,3 +190,64 @@ func (s *KeywordDBTestSuite) TestGetKeywordsByInvalidColumn() {
...
+
+	newKeyword := Keyword{Keyword: "Ronaldo"}
+
+	invalidKeywordID := 999999
+	err := UpdateKeyword(uint(invalidKeywordID), newKeyword)
+
+	var result Keyword
+	db.GetDB().First(&result, keyword.ID)
+
+	assert.Equal(s.T(), nil, err)
```
andyduong1920, line 239: 🤔 why the error here is `nil` it should be not nil` no?

andyduong1920: something like Not found instead IMO or we would like to keep this as a silent error 🤔

services/google_search_service/db_updater.go
```diff
@@ -0,0 +1,53 @@
+package google_search_service
+
+import (
+	"encoding/json"
+
+	"github.com/gutakk/go-google-scraper/models"
+)
+
+func UpdateKeywordStatus(keywordID uint, status models.KeywordStatus, failedReason error) error {
+	var keywordModel models.Keyword
```
andyduong1920: how about just call it as `keyword` ☁️

services/google_search_service/db_updater.go
```diff
@@ -0,0 +1,53 @@
...
+	if nonAdwordLinksParsingErr != nil {
+		return nonAdwordLinksParsingErr
+	}
+
+	topPositionAdwordLinks, topPositionAdwordLinksParsingErr := json.Marshal(parsingResult.TopPositionAdwordLinks)
+	if topPositionAdwordLinksParsingErr != nil {
+		return topPositionAdwordLinksParsingErr
+	}
+
+	newKeyword := models.Keyword{
```
andyduong1920: just keyword is ok IMO

services/google_search_service/job_enqueuer.go
```diff
@@ -0,0 +1,38 @@
+package google_search_service
+
+import (
+	"errors"
+	"log"
+
+	"github.com/gocraft/work"
+	"github.com/gutakk/go-google-scraper/db"
```
andyduong1920, line 7: as `"github.com/gutakk/go-google-scraper/db"` is our internal lib and the `"github.com/gocraft/work"` is external, should we add a `line break` between these group @gutakk 🤔

services/google_search_service/job_enqueuer_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"github.com/gocraft/work"
+	"github.com/gomodule/redigo/redis"
+	"github.com/gutakk/go-google-scraper/db"
+	"github.com/gutakk/go-google-scraper/models"
+	testDB "github.com/gutakk/go-google-scraper/tests/db"
+
+	"github.com/stretchr/testify/suite"
+	"gopkg.in/go-playground/assert.v1"
```
andyduong1920, line 14: I'm not sure about the order here, as I understand the `"github.com/gocraft/work"`, `"github.com/gomodule/redigo/redis"` also the external lib, should we move it below @gutakk

models/keyword_test.go
gutakk: I'm not sure about another ORM or RAW sql. As I using `GORM` this scenario doesn't return an error

this is my `UpdateKeyword` function
```go
func UpdateKeyword(keywordID uint, newKeyword Keyword) error {
	result := db.GetDB().Model(&Keyword{}).Where("id = ?", keywordID).Updates(newKeyword)
	if result.Error != nil {
		return result.Error
	}

	return nil
}
```

Do you have any idea?

services/google_search_service/db_updater.go
gutakk: Fixed in 36d1470f8bb708c6c8e20c3cffd9deff9add9370

services/google_search_service/db_updater.go
gutakk: Fixed in 36d1470f8bb708c6c8e20c3cffd9deff9add9370

services/google_search_service/job_enqueuer.go
gutakk: Good catch. It was my sloppy. Fixed in 162033b0a7ad6424a236310201baad3a5cc97003

services/google_search_service/job_enqueuer_test.go
gutakk: Fixed in 162033b0a7ad6424a236310201baad3a5cc97003

models/keyword_test.go
andyduong1920: If it's the default GORM then we could use this, for now, @gutakk

services/google_search_service/db_updater_test.go
```diff
@@ -0,0 +1,106 @@
+package google_search_service
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/bxcodec/faker/v3"
```
andyduong1920: Small thing, `github.com/bxcodec/faker/v3` is alo an external lib, is it correct @gutakk 😀

gutakk: Exactly, fixed in f7c2c7bd294695117f9990e97ec4d91c96eebe35.
For another old files I will refactor to follow our convention in https://github.com/gutakk/go-google-scraper/issues/33

andyduong1920 (approved):
🎉

PR #31, opened 2020-12-15
As a user, I can see scraping status

### LINKED ISSUE #30: As a user, I can see scraping status
### Acceptance Criteria
- UI to display scraping status

gutakk:
## What happened
- [x] Add status to keyword list table
- [x] A little refactor

## Insight
- [x] Move formatted createdAt from models to presenters

## Proof Of Work
- Created at display properly and status display differently depend on value
Screen Shot 2563-12-16 at 10 22 43


presenters/keyword.go
```diff
@@ -0,0 +1,7 @@
+package presenters
+
+import "time"
+
+func FormattedCreatedAt(createdAt time.Time) string {
```
olivierobert: Good you decided to improve on this 💪 But this look more like a view helper 🙈 

An actual presenter would take a keyword as argument:

```go
package presenters

import (
  "time"

  "github.com/gutakk/go-google-scraper/models"
)

type KeywordPresenter struct {
  Keyword models.Keyword
}

func (kp *KeywordPresenter) FormattedCreatedAt() string {
  return kp.Keyword.createdAt.Format("January 2, 2006")
}
```

So in that sense, the `KeywordPresenter` would only be initialized in the `KeywordController`

gutakk: I think I would go with view helper instead of presenter in this case because this controller need `[]models.Keyword` and display all of them in the view. It is quite difficult to implement the presenter to store data type like this.

What do you think?

olivierobert: It's not that hard actually 😁 In the controller, you would need to pass a slide of keyword presenters. Instead of accessing the keyword model directly, the view would access the keyword model via keyword presenter. That's how it's done in Rails projects as well.

View helpers should be avoided in most cases as they become hard to maintain over time.

gutakk: Fixed in ae15286cd56a4f868b9ca2687846005d7b0e1e84

olivierobert (approved):

andyduong1920 (approved):

PR #32, opened 2020-12-16
As a user, I can see scraping result

### LINKED ISSUE #29: As a user, I can see scraping result
### Acceptance Criteria
- UI to display scraping result
- Include test

### LINKED ISSUE #30: As a user, I can see scraping status

### LINKED ISSUE #37: Worker cannot run locally due to unknown version of AIR
## Issue

When trying to run the application from the branch `feature/scraping-result-page`, it failed:

image

So I tried to re-set up the dependencies with `make build-dependencies` bit failed for the same reasons.

image
 
## Expected

It might be a case of version disappearing on Github, so the dependency version should be updated.
 
## Steps to reproduce

1. Clone the branch `feature/scraping-result-page`
2. Start the application with `make start-dev`

gutakk:
## What happened
Resolve #29 

## Insight
- [x] Add keyword result page with style
- [x] Add google html page
- [x] Add 404 not found page
  - Invalid keyword result page
  - Invalid google html page
  - Invalid url
- [x] Remove nav bar as this will be implemented in https://github.com/gutakk/go-google-scraper/issues/35
- [x] Add presenter for keyword result page

## Proof Of Work
- Keyword list page (add action link)
Screen Shot 2563-12-18 at 18 26 08

- Keyword result page
Screen Shot 2563-12-18 at 18 27 48

- Google HTML page (click at google icon)
Screen Shot 2563-12-18 at 18 28 12

- Not found page
Screen Shot 2563-12-18 at 18 28 35


assets/stylesheets/components/_search-result-list.scss
```diff
@@ -0,0 +1,23 @@
+.search-result-list {
```
olivierobert: > Use the UI element type as a prefix following by one to two words. While it might seem counter-intuitive at first, this technique makes it easier to differentiate component purpose and search through components. Most framework use this technique e.g. btn-primary is not called primary-btn as btn is the namespace grouping all buttons.
> https://nimblehq.co/compass/development/code-conventions/css/#class-names

Prefer naming it `.list-search-result`

templates/keyword.html
```diff
@@ -31,7 +31,13 @@ <h2>Upload CSV</h2>
               <td><strong>{{ $k.Keyword }}</strong></td>
               <td>{{ $k.FormattedCreatedAt }}</td>
               <td></td>
-              <td></td>
+              <td>
+                {{ if eq $k.Status "processed" }}
+                  <a href="/keyword/{{ $k.ID }}">
+                    <img src="/dist/svg/document.svg">
```
olivierobert, line 39: Some issues with this implementation:
- Performance: this SVG image would need to be rendered in the DOM as many time as there are rows in the table. Besides, this image would be fetched very late when rendering the first row.
- Accessibility: this image would not be rendered by screen readers

Instead, prefer using CSS background for that:

```html
<a class="list-keyword__status list-keyword__status--processed" href="/keyword/{{ $k.ID }}"><span class="sr-only">Processed</span></a>
```

> See https://tailwindcss.com/docs/screen-readers

WHen using inline SVG images, prefer to create a sprite image as well.

templates/not_found.html
```diff
@@ -0,0 +1,17 @@
...
+
+<html>
+  <head>
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <link type="text/css" rel="stylesheet" href="/dist/stylesheets/index.css" />
+    <title>Not found</title>
+  </head>
+
+  <body class="not-found">
```
olivierobert: Never use `body` as the Block (B in BEM), instead prefer to create a component `div`:

```
<body>
  <div class="not-found">

  </div>
</body>
```

Also, it might be better to create a layout later on for these error pages 💭 

```
<HTML class="layout-error">
```

services/keyword_service/keyword_test.go
```diff
@@ -86,6 +86,47 @@ func (s *KeywordServiceDbTestSuite) TestGetAllWithInvalidUser() {
...
+}
+
+func (s *KeywordServiceDbTestSuite) TestGetKeywordResultWithWrongKeywordTypeButValidUser() {
+	keyword := models.Keyword{UserID: s.userID, Keyword: faker.Name()}
+	db.GetDB().Create(&keyword)
+
+	result, err := s.keywordService.GetKeywordResult("invalidKeyword")
+
+	assert.Equal(s.T(), models.Keyword{}, result)
+	assert.Equal(s.T(), "ERROR: invalid input syntax for type bigint: \"invalidKeyword\" (SQLSTATE 22P02)", err.Error())
```
olivierobert: This kind of error messages are really not user friendly 🙈 No way with GORM to have a better message like "keyword not found for ..." Or handle it at the service level.

Saem above for `TestGetKeywordResultWithValidKeywordButInvalidUser` as it should just return a similar error

templates/partials/search_result_list.html
```diff
@@ -0,0 +1,21 @@
...
+
+<div class="search-result-list">
+  <dt class="search-result-list__title">
+    {{ .title }}
+  </dt>
+  {{ if .value }}
+    <ul class="search-result-list__value divide-y divide-gray-200">
+      {{ range $v := .value }}
+        <li class="search-result-list__item">
+          <img class="search-result-list__icon" src="/dist/svg/link.svg">
```
olivierobert: Same comment as above, prefer using CSS instead of inlining images. Inlining SVGS should be reserved for cases when it's necessary to override the path colours or strokes based on CSS classes. In most other cases, prefer using a background image.

olivierobert (commented):
I could not check the frontend UI in details just yet as I faced an issue: https://github.com/gutakk/go-google-scraper/issues/37

assets/stylesheets/components/_search-result-list.scss
gutakk: Fixed in 79776e2101239b7c67f5b2b7ac8a01ddb7ec5822

templates/not_found.html
gutakk: Fixed in 4fedf6614da6e5887fe0be91dad360de42998134

services/keyword_service/keyword_test.go
gutakk: Fixed in c08a6c9c7ba927707ba5830bf53c8a1078368a55

templates/keyword.html
gutakk: Decide to do this in another PR as I found some issue with ParcelJS
issue link: https://github.com/gutakk/go-google-scraper/issues/41

templates/partials/search_result_list.html
gutakk: Decide to do this in another PR as I found some issue with ParcelJS
issue link: https://github.com/gutakk/go-google-scraper/issues/41

helpers/error_handler/error_message.go
```diff
@@ -51,6 +53,8 @@ func DatabaseErrorMessage(err error) error {
 			return errors.New(somethingWentWrongError)
 		case duplicateErrorCode:
 			return errors.New(emailDuplicateError)
+		case invalidInputErrorCode:
```
olivierobert, line 58: Instead of checking on `pgerr.code`, it migth be better to use `pgerr.Code.Class().Name()`:

```go
switch pgerr.Code.Class().Name() {
case "data_exception":
  err = ErrInvalidValue
case "integrity_constraint_violation":
  // This could be lots of things: foreign key violation,
  // non-null constraint violation, etc., but we're generally
  // checking those in advance. As long as our code is in
  // order, unique constraints will be the only things we're
  // actually relying on the database to check for us.
  err = ErrDuplicate
}
```

Here is the list of values: https://github.com/lib/pq/blob/master/error.go#L78

You can see more info on [this blog post](https://ders.github.io/post/2016-12-23-lean-clean-golang-machine/).

What do you think?

Also, there seem to be an attribute `Message` and `Detail`:

https://github.com/lib/pq/blob/4604d39ddc9f62e2ca152114c73aec99b77a3468/error.go#L28-L29

So wondering if this could be used to have better error messages too.

controllers/keyword.go
```diff
@@ -28,6 +30,8 @@ type UploadFileForm struct {
 
 func (k *KeywordController) applyRoutes(engine *gin.RouterGroup) {
 	engine.GET("/keyword", k.displayKeyword)
+	engine.GET("/keyword/:keyword_id", k.displayKeywordResult)
+	engine.GET("/keyword/:keyword_id/google-html", k.displayKeywordGoogleHTML)
```
olivierobert: How about simplifying it to `/keyword/:keyword_id/html` as it would be flexiblee if other search engines are supported.

helpers/error_handler/error_message.go
```diff
@@ -11,13 +11,15 @@ const (
 	commonFieldError        = "%s is not valid"
 	emailFormatError        = "invalid email format"
 	emailDuplicateError     = "email already exists"
+	invalidInputError       = "invalid input"
```
olivierobert, line 15: This message is still too generic and not very helpful 😵 It can work for now but I am still trying to find better ways to handle errors with form. See my comment below.

controllers/keyword.go
gutakk: Good catch. Fixed in 01ba0af024234cbefe807cc480b62eceab3eff80

helpers/error_handler/error_message.go
gutakk: `GORM` doesn't use `https://github.com/lib/pq` but they use `github.com/jackc/pgconn` so `Class()` and `Name()` functions are not available.

pgconn also provide `Message` and `Detail` attributes and this is the result
- Message : `invalid input syntax for type bigint: "invalidKeyword"`
- Detail : empty string

for more information [click](https://github.com/jackc/pgconn/blob/master/errors.go)

helpers/error_handler/error_message.go
gutakk: Replied [here](https://github.com/gutakk/go-google-scraper/pull/32#discussion_r548437819)

assets/stylesheets/index.scss
```diff
@@ -6,16 +6,19 @@
...
 @import 'components/alert';
+@import 'components/search-result';
+@import 'components/search-result-list';
 
 // Layouts
 @import 'layouts/application';
 @import 'layouts/authentication';
 
 // Screens
+@import 'screens/keyword_result.scss';
```
andyduong1920: do we need to add the `.scss` extension @gutakk 🤔

services/keyword_service/keyword_test.go
```diff
@@ -87,6 +87,47 @@ func (s *KeywordServiceDbTestSuite) TestGetAllWithInvalidUser() {
...
+
+func (s *KeywordServiceDbTestSuite) TestGetKeywordResultWithValidKeywordButInvalidUser() {
+	keyword := models.Keyword{UserID: s.userID, Keyword: faker.Name()}
+	db.GetDB().Create(&keyword)
+
+	keywordService := KeywordService{CurrentUserID: 999999}
+	result, err := keywordService.GetKeywordResult(keyword.ID)
+
+	assert.Equal(s.T(), models.Keyword{}, result)
+	assert.Equal(s.T(), "something went wrong, please try again", err.Error())
```
andyduong1920: Should we change the message into `Unauthorize` as `something went wrong, please try again` is not really correct for this case @gutakk

assets/stylesheets/index.scss
gutakk: Good catch. Fixed in 21657bced92265bf53ea7fe908092eef633233f9

services/keyword_service/keyword_test.go
gutakk: IMO, it is not `Unauthorize` but it should be something like `keyword not found` because this is the function that query from database and it is not related to authentication

so I fixed the error message from `something went wrong, please try again` to `keyword not found` in 406cbbdb8458be74464f4342095d1f2458956f9f

helpers/error_handler/error_message.go
olivierobert: Thank you for the detailed info. Still not helping us to get better error messages 😅 but now I have more info on my hand.

assets/stylesheets/screens/not_found.scss
```diff
@@ -0,0 +1,19 @@
+.not-found {
```
olivierobert: Missing the body tag here:

```suggestion
body.not-found {
```

presenters/keyword.go
```diff
@@ -11,3 +14,39 @@ type KeywordPresenter struct {
 func (kp *KeywordPresenter) FormattedCreatedAt() string {
 	return kp.Keyword.CreatedAt.Format("January 2, 2006")
 }
+
+type KeywordResult struct {
```
olivierobert: With some hindsight having this struct defined below the other method seems odd 🤔 Also `KeywordResult` seems to duplicate entirely the Keyword struct which is not good for maintainability in the sense that adding an attribute in the model struct would require to modify the presenter struct here as well. 

Note that all attributes just as ID and Status are not altered to the view can still access those as: `kp.Keyword.ID` or `kp.Keyword.Status`. So there is no need to do `kp.KeywordResult.ID` 🙃 Only `NonAdwordLinks` and `topPositionAdwordLinks` are actually formatted here.

olivierobert (changes requested):

assets/stylesheets/screens/not_found.scss
gutakk: Fixed in 3853dc031eddc5caad6b6194e9887f20a3332492

presenters/keyword.go
gutakk: Agree with you. Fixed in 803f92c3c436caec1a94f017310272c8e74d8820

presenters/keyword.go
```diff
@@ -1,13 +1,33 @@
...
 )
 
 type KeywordPresenter struct {
 	Keyword models.Keyword
 }
 
+type KeywordResult struct {
+	NonAdwordLinks         []string
+	TopPositionAdwordLinks []string
+}
```
olivierobert, line 16: Thanks for making the adjustment to simplify this struct. But I still think this struct can be removed altogether. I am not sure why you preferred to have a struct instead of two method like:

```
func (kp *KeywordPresenter) FormattedNonAdwordLinks() []string {
	var nonAdwordLinks []string
	return json.Unmarshal(kp.Keyword.NonAdwordLinks, &nonAdwordLinks)
}
```

In the view, it would prevent switching between `keyword` and `keywordResult`:

```
    {{ template "search_result" searchResultPartial "Total Non Adword Links" .data.keyword.NonAdwordsCount }}
    {{ template "search_result_list"  searchResultPartial "Non Adword Links" .data.keywordResult.NonAdwordLinks }}
```

What I recommend to is to simply pass the presenter only:

```
    {{ template "search_result" searchResultPartial "Total Non Adword Links" .data.keywordPresenter.keyword.NonAdwordsCount }}
    {{ template "search_result_list"  searchResultPartial "Non Adword Links" .data.keywordPresenter.FormattedNonAdwordLinks }}
```
```

Comment (gutakk) [reply presenters/keyword.go:16]: That is the good way to go but that solution lead to this [issue](https://github.com/gutakk/go-google-scraper/issues/46)

When I use like this
```html
{{ template "search_result_list"  searchResultPartial "Non Adword Links" .data.keywordPresenter.FormattedNonAdwordLinks }}
```
this is an error
```
executing "content" at <.data.keywordPresenter.FormattedTopPositionAdwordLinks>: can't evaluate field FormattedTopPositionAdwordLinks in type interface {}
```

After I researched on google and try this
```html
{{ with .data.keywordPresenter }}
    {{ template "search_result_list"  searchResultPartial "Top Position Adword Links" .FormattedTopPositionAdwordLinks }}
{{ end }}
```

and this is an error
```
executing "content" at <.FormattedTopPositionAdwordLinks>: can't evaluate field FormattedTopPositionAdwordLinks in type presenters.KeywordPresenter
```

I still can't find the solution how to solve this

andyduong1920 (approved):

assets/stylesheets/index.scss
```diff
@@ -6,16 +6,19 @@
...
 @import 'base/button';
 @import 'base/form';
-@import 'base/nav';
 @import 'base/table';
 @import 'base/typography';
 
 // Components
 @import 'components/alert';
+@import 'components/search-result';
+@import 'components/search-result-list';
```
olivierobert: Since the class name is `.list-search-result`, teh file name must be:

```suggestion
@import 'components/list-search-result';
```

assets/stylesheets/index.scss
```diff
@@ -6,16 +6,19 @@
...
 
 // Layouts
 @import 'layouts/application';
 @import 'layouts/authentication';
 
 // Screens
-@import 'screens/keyword.scss';
+@import 'screens/keyword_result';
```
olivierobert: Use hyphens for file names:

```suggestion
@import 'screens/keyword-result';
```
Same issue for `'screens/not_found';` -> `'screens/not-found';

templates/not_found.html
olivierobert: That fix does not cover all the requirements as the html element still does not have any class name.

Since these pages have a different layout, prefer to create a new layout as advised before:

```
<html class="layout-error">
```

Then create new base styles for all error pages. It can be reused for 401 errors when a user is not authenticated and 403 when permissions are not sufficient.

The content can be left simple:

```
<body class="not-found">
  <div class="error-details">
    <h2 class="error-details__title"> 404 </h2>
    ...
  </div>
</body>
```

Currently, there are many issues:
- `<main>` is slightly overkill as there is no other content on the page

image


- Never use HTML tags in the class naming so `main-not-found` violates [this convention](https://nimblehq.co/compass/development/code-conventions/css/#class-names)

templates/keyword_result.html
```diff
@@ -0,0 +1,30 @@
+{{ define "content" }}
+
+<div class="result-keyword">
```
olivierobert: How about making it `<section>`

```suggestion
<section class="result-keyword">
```

So that below we can use the header tag:

```
<header class="result-keyword__header">
```

templates/keyword_result.html
```diff
@@ -0,0 +1,30 @@
...
+  <div class="result-keyword__header">
+    <a class="result-keyword__google-link" href="/keyword/{{ .data.keyword.ID }}/html" target="_blank">
+      <img src="/dist/svg/google.svg">
```
olivierobert, line 7: You are not using the link alt attribute and image replacement technique we discussed before (using `<span class="sr-only">View the cached HTML page</span>`, see the [Tailwind Docs](https://tailwindcss.com/docs/screen-readers)). 

HTML must be semantic and accessible 💪

presenters/keyword.go
olivierobert: Noted. How about then using a more meaningful name:

```
type KeywordLinks struct {
    NonAdwordLinks         []string
	TopPositionAdwordLinks []string
}
```

`KeywordResult` is neither very clear or explicit.

templates/partials/search_result_list.html
```diff
@@ -0,0 +1,21 @@
+{{ define "search_result_list" }}
+
+<div class="list-search-result">
+  <dt class="list-search-result__title">
+    {{ .title }}
+  </dt>
+  {{ if .value }}
+    <ul class="list-search-result__value divide-y divide-gray-200">
```
olivierobert: When using `dt`, the companion tag is `<dd>` as used in `templates/partials/search_result.html`:

```suggestion
    <dd class="list-search-result__value">
      <ul>
```

Also not sure why the styles `divide-y divide-gray-200` are not in the stylesheets 🤔  Unless this is a variation from the base styles, having them all in one place would make more sense.

olivierobert (changes requested):

templates/not_found.html
gutakk: Thanks for clarify this.
Fixed in 1295119899b58fc9ccdb3353ff7e7bcecbaaed24

presenters/keyword.go
gutakk: Fixed in 678cf1436697f08804df0e0f675fb80568a697cd

assets/stylesheets/index.scss
gutakk: Fixed in f8bb9f26308ae2bbe6c5ec94a948f49bea25862b

assets/stylesheets/index.scss
gutakk: Good catch! Fixed in 5e3a7d09fca8378612cee9a70a962b78c6f12e93

templates/keyword_result.html
gutakk: Nice idea. Fixed in 60a0a30250d5fd6437c773d0514e0347087250d5 d9ad685bde7b9780139fa31e19ff43772b488207

templates/keyword_result.html
gutakk: Decide to do this in another PR as I found some issue with ParcelJS
issue link: #41

templates/partials/search_result_list.html
gutakk: Added `<dd>` in 7e5236d72de5de5ab0932b3ecf40ecaf0ed61e6a

gutakk: For `divide-y divide-gray-200`, they cannot apply to stylesheets because tailwind version that lower than 2 doesn't support `@apply` for the class that combine more than one styles.

- `divide-y`
   ```css
   --tw-divide-y-reverse: 0;
   border-top-width: calc(1px * calc(1 - var(--tw-divide-y-reverse)));
   border-bottom-width: calc(1px * var(--tw-divide-y-reverse));
   ```

- `divide-gray-200`
   ```css
   --tw-divide-opacity: 1;
   border-color: rgba(229, 231, 235, var(--tw-divide-opacity));
   ```

Tailwind version above 2 has an issue with the ParcelJS that's why I use version lower than 2. Not sure if they fix that issue already.

templates/keyword_result.html
olivierobert: Oh. it's still the same issue as before 😔 I am fine to solve this in another PR but do create a TODO in the codebase to avoid others or me 😓 to raise this issue again.

Also in terms of UX, I don't think most users would understand that clicking on the "G" icon would open the HTML page 🤔 That's why having a link would make more sense.

image

templates/partials/search_result_list.html
olivierobert: When was the last time you check the issue with Parcel JS and Tailwind 2. I found several tutorials like [this one](https://tailwind.club/setting-up-tailwind-v2-with-parcel/) that shows that it should work. I am fine to solve this in another PR but do create a TODO in the codebase to avoid others or me 😓 to raise this issue again.

presenters/keyword_test.go
```diff
@@ -21,3 +24,25 @@ func TestFormattedCreatedAtWithValidTime(t *testing.T) {
...
+		NonAdwordLinks:          nonAdwordLinks,
+		TopPositionAdwordsCount: 10,
+		TopPositionAdwordLinks:  topPositionAdwordLinks,
+		TotalAdwordsCount:       10,
+		HtmlCode:                "test-html",
+	}
+
+	keywordPresenter := KeywordPresenter{Keyword: keyword}
+	result := keywordPresenter.KeywordLinks()
+
```
olivierobert, line 44: To be more complete, this test should also test every single attribute:

```
assert.Equal(t, "test-keyword", keywordPresenter.Keyword.Keyword)
assert.Equal(t, "pending", keywordPresenter.Keyword.Status)
#...
```

The fact the keyword model can be accessed through the KeywordPresenter class is a feature.

That's why in the setup step, here using `faker.Name()` is not the right decision to make assertions:

```
keyword := models.Keyword{
  Keyword: "test-keyword",
```

templates/keyword_result.html
gutakk: Added TODO in 0245d45c0d3d5de938135367e2dab1eb5e05c017

templates/partials/search_result_list.html
gutakk: The last time I checked is about 2 months ago.
Added TODO in 09eb29c0c55cf15d0808ea4ce3402c02694e2b0d

assets/stylesheets/components/_list-search-result.scss
```diff
@@ -0,0 +1,23 @@
...
+    @apply text-sm font-medium text-gray-500;
+  }
+
+  &__value {
+    @apply border border-gray-200 rounded-md col-span-2;
+  }
+
+  &__item {
+    @apply pl-3 pr-4 py-3 flex items-center text-sm;
+  }
```
carryall: @gutakk How about like this?
```suggestion
  &__title,
  &__item {
    @apply text-sm;
  }

  &__title {
    @apply font-medium text-gray-500;
  }

  &__value {
    @apply border border-gray-200 rounded-md col-span-2;
  }

  &__item {
    @apply pl-3 pr-4 py-3 flex items-center;
  }
```

templates/keyword_result.html
gutakk: Added link under the G icon in 392c9a7d0d6407b983884f07421656e2e6deed0b
Screen Shot 2564-01-20 at 16 00 05

assets/stylesheets/components/_search-result.scss
```diff
@@ -0,0 +1,11 @@
+.search-result {
+  @apply px-6 py-5 grid grid-cols-3 gap-4;
+
+  &__title {
+    @apply text-sm font-medium text-gray-500;
+  }
+
+  &__value {
+    @apply mt-0 text-sm text-gray-900 col-span-2;
+  }
```
carryall:
```suggestion
  &__title,
  &__value {
    @apply text-sm;
  }

  &__title {
    @apply font-medium text-gray-500;
  }

  &__value {
    @apply mt-0 text-gray-900 col-span-2;
  }
```

presenters/keyword_test.go
gutakk: Updated in 9443497a28ae6a66460d0b21de9655626d82df83

assets/stylesheets/components/_list-search-result.scss
gutakk: Looks great. Fixed in eb61320c6b7dcf1d9baa7c548c2f4e0707abb9c2

assets/stylesheets/components/_search-result.scss
gutakk: Fixed in 603340805ae9f64933a351c7dbf84bd1a98f7ef5

services/keyword_service/keyword_test.go
```diff
@@ -87,6 +87,47 @@ func (s *KeywordServiceDbTestSuite) TestGetAllWithInvalidUser() {
...
+
+	assert.Equal(s.T(), models.Keyword{}, result)
+	assert.Equal(s.T(), "invalid input", err.Error())
+}
+
+func (s *KeywordServiceDbTestSuite) TestGetKeywordResultWithInvalidKeywordButValidUser() {
+	keyword := models.Keyword{UserID: s.userID, Keyword: faker.Name()}
+	db.GetDB().Create(&keyword)
+
+	result, err := s.keywordService.GetKeywordResult("999999")
```
carryall, line 125: @gutakk Would the result be the same if we change `"999999"` to integer `999999`? 🤔 
Also how about changing the function name to `TestGetKeywordResultWithInvalidKeywordIDButValidUser` because we are passing the invalid keyword ID.

assets/stylesheets/index.scss
```diff
@@ -6,16 +6,19 @@
...
+@import 'components/list-search-result';
+@import 'components/search-result';
 
 // Layouts
 @import 'layouts/application';
 @import 'layouts/authentication';
 
 // Screens
-@import 'screens/keyword.scss';
+@import 'screens/keyword-result';
```
hoangmirs, line 22: As we import these screens here, should we rename them with `_` prefix following [our convention](https://nimblehq.co/compass/development/code-conventions/css/#files)

services/keyword_service/keyword_test.go
gutakk: Good catch for the function name. Fixed in 641accbf6d741e60b2c80e4583ed186f9e148c22

The `GetKeywordResult` parameters type is `interface{}` and it can take both string and integer and the result is the same. I added more test case about it in 3c4c3f622a6f2b5747b140097012b9bfbab3a81b

assets/stylesheets/index.scss
gutakk: Good catch. Fixed in 2094674afe24e1f6e6bd1e01c0bc1a7ac7a32460

olivierobert: The fix still does not show though 🤔 

image

This would apply to all imported stylesheets 🙈 So maybe doing this clean up in a chore task can be the good call here.

olivierobert (approved):

hoangmirs (approved):

assets/stylesheets/index.scss
gutakk: Created issue for this change in https://github.com/gutakk/go-google-scraper/issues/66

carryall (approved):

Issue #41, opened 2020-12-22
Update SVGs usage


gutakk:
Some issues with this implementation:
- Performance: this SVG image would need to be rendered in the DOM as many time as there are rows in the table. Besides, this image would be fetched very late when rendering the first row.
- Accessibility: this image would not be rendered by screen readers

Instead, prefer using CSS background for that:

```html
<a class="list-keyword__status list-keyword__status--processed" href="/keyword/{{ $k.ID }}"><span class="sr-only">Processed</span></a>
```

> See https://tailwindcss.com/docs/screen-readers

WHen using inline SVG images, prefer to create a sprite image as well.

_Originally posted by @olivierobert in https://github.com/gutakk/go-google-scraper/pull/32#discussion_r546559046_

and

This is another example of when a background image should be used instead:

```html
<div class="navigation__logo">
  <a href="/"><span class="sr-only>">Go Google Scraper</span></a>
</div>
```

> `sr-only` is for screen-reader so for accessibility

_Originally posted by @olivierobert in https://github.com/gutakk/go-google-scraper/pull/42#discussion_r548374406_

PR #42, opened 2020-12-23
As a user, I can navigate through application

### LINKED ISSUE #35: As a user, I can navigate through application
## Acceptance Criteria
- Add the navbar with style

gutakk:
## What happened
Resolved #35 

## Insight
- [x] Add nav bar at the top of every screen
- [x] Nav bar show content as following
  - Authenticated user: Keyword and Logout
  - Guest user: Login and Register

## Proof Of Work
- Authenticated user nav bar
Screen Shot 2563-12-23 at 17 26 29

- Guest user nav bar
Screen Shot 2563-12-23 at 17 26 58

### DIFF
--- a/services/keyword_service/keyword.go
+++ b/services/keyword_service/keyword.go
@@ -25,12 +25,12 @@ const (
 )
 
 type KeywordService struct {
-	CurrentUser models.User
+	CurrentUserID uint
 }
 
 func (k *KeywordService) GetAll() ([]models.Keyword, error) {
 	condition := make(map[string]interface{})
-	condition["user_id"] = k.CurrentUser.ID
+	condition["user_id"] = k.CurrentUserID
 
 	keywords, err := models.GetKeywordsBy(condition)
 	if err != nil {
@@ -43,7 +43,7 @@ func (k *KeywordService) GetAll() ([]models.Keyword, error) {
 func (k *KeywordService) GetKeywordResult(keywordID interface{}) (models.Keyword, error) {
 	condition := make(map[string]interface{})
 	condition["id"] = keywordID
-	condition["user_id"] = k.CurrentUser.ID
+	condition["user_id"] = k.CurrentUserID
 
 	keyword, err := models.GetKeywordBy(condition)
 	if err != nil {
@@ -60,7 +60,7 @@ func (k *KeywordService) Save(parsedKeywordList []string) error {
 	}
 
 	for _, value := range parsedKeywordList {
-		keyword := models.Keyword{Keyword: value, UserID: k.CurrentUser.ID}
+		keyword := models.Keyword{Keyword: value, UserID: k.CurrentUserID}
 
 		txErr := db.GetDB().Transaction(func(tx *gorm.DB) error {
 			savedKeyword, err := models.SaveKeyword(keyword, tx)
--- a/services/keyword_service/keyword.go
+++ b/services/keyword_service/keyword.go
@@ -25,12 +25,12 @@ const (
 )
 
 type KeywordService struct {
-	CurrentUserID uint
+	CurrentUser models.User
 }
 
 func (k *KeywordService) GetAll() ([]models.Keyword, error) {
 	condition := make(map[string]interface{})
-	condition["user_id"] = k.CurrentUserID
+	condition["user_id"] = k.CurrentUser.ID
 
 	keywords, err := models.GetKeywordsBy(condition)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


services/keyword_service/keyword.go
```diff
@@ -43,7 +43,7 @@ func (k *KeywordService) GetAll() ([]models.Keyword, error) {
 func (k *KeywordService) GetKeywordResult(keywordID interface{}) (models.Keyword, error) {
 	condition := make(map[string]interface{})
 	condition["id"] = keywordID
-	condition["user_id"] = k.CurrentUserID
+	condition["user_id"] = k.CurrentUser.ID
```
olivierobert: That's authorization done on the cheap 😆 But it works

services/keyword_service/keyword_test.go
```diff
@@ -101,7 +105,11 @@ func (s *KeywordServiceDbTestSuite) TestGetKeywordResultWithValidKeywordButInval
 	keyword := models.Keyword{UserID: s.userID, Keyword: faker.Name()}
 	db.GetDB().Create(&keyword)
 
-	keywordService := KeywordService{CurrentUserID: 999999}
+	keywordService := KeywordService{
+		CurrentUser: models.User{
+			Model: gorm.Model{ID: 999999},
```
olivierobert: This syntax is quite odd. Can you please explain why you went this way to create a User fixture 🤔

templates/layouts/application.html
```diff
@@ -9,12 +9,14 @@
   </head>
 
   <body class="{{ .view }}">
+    {{ include "partials/nav" }}
+
     <main class="app-main">
-      <div class="app-main__content">
+    <div class="app-main__content">
```
olivierobert: Missing indentation

templates/partials/nav.html
```diff
@@ -0,0 +1,19 @@
+<nav class="navigation">
+  <div>
+    <a href="/"><img class="navigation__logo" src="/dist/images/google-go.jpg"></img></a>
```
olivierobert, line 3: This is another example of when a background image should be used instead:

```html
<div class="navigation__logo">
  <a href="/"><span class="sr-only>">Go Google Scraper</span></a>
</div>
```

> `sr-only` is for screen-reader so for accessibility

templates/partials/nav.html
```diff
@@ -0,0 +1,19 @@
...
+  </div>
+
+  {{ if .data.authenticatedUser }}
```
olivierobert, line 6: That's where a presenter would be better to group all view related methods:

```
{{ if .data.applicationPresenter.userIsSignedIn }}
```

Now there is only one authenticated controller but if there are more than one then not having this code in the keyword controller only would be necessary.

templates/partials/nav.html
```diff
@@ -0,0 +1,19 @@
...
+    <div>
+      <a class="navigation__link" href="/keyword">Keyword</a>
+      <form class="navigation__form" action="/logout" method="POST">
+        <input class="navigation__input" type="submit" value="Logout"></input>
```
olivierobert: Instead of `input`, that should be a button:

```suggestion
        <button class="navigation__button" type="submit">Logout</button>
```

services/keyword_service/keyword_test.go
gutakk: Because `KeywordService` struct take only `CurrentUserID` which is `uint` type in the first place but right now I decided to take the whole `User` model (to display email and use ID to check authentication) and the `ID` came from gorm Model struct.

gutakk: But as I change the design, keyword page doesn't need to display email anymore. I will revert back to previous version.

gutakk: Reversed in cc367782840cad65293683843acfc26beb759628

templates/layouts/application.html
gutakk: Fixed in ede28fe6bab73e05d5ee2d0535318a7030402edb

templates/partials/nav.html
gutakk: I will update the usage of images and svgs in this [issue](https://github.com/gutakk/go-google-scraper/issues/41)

templates/partials/nav.html
gutakk: Fixed in 54a708cb0c2aad215551516f779c5e54293e16f9

templates/partials/nav.html
gutakk: I found issue with the gin context. I cannot use function from the presenter in the view. So I will do this in another PR
https://github.com/gutakk/go-google-scraper/issues/46

andyduong1920 (approved):

olivierobert (approved):

olivierobert:
@gutakk Rebased needed so that it can be merged

hoangmirs (approved):

PR #44, opened 2020-12-23
As a user, I can see nice landing page

### LINKED ISSUE #43: As a user, I want to see nice landing page.
## Acceptance Criteria
- Style home page

gutakk:
## What happened
Resolved #43 

## Insight
- [x] Style home page

## Proof Of Work
- Home page
Screen Shot 2563-12-23 at 18 10 20


assets/stylesheets/screens/home.scss
```diff
@@ -0,0 +1,16 @@
+.home {
```
olivierobert: Bind it to the `body` tag to avoid with conflicts with a component named `home` for instance 💥 :

```suggestion
body.home {
```

make sure this is done for all screen stylesheets.

templates/home.html
```diff
@@ -1,17 +1,18 @@
...
-    <button type="submit">Logout</button>
-  </form>
-{{ else }}
-  <h2><a href="/login">Login</a></h2>
-  <h2><a href="/register">Register</a></h2>
-{{ end }}
+  <div class="home">
+    {{ include "partials/flashes" }}
+    
+    <p class="home__title sm:text-5xl md:text-6xl">
```
olivierobert: How about using a heading instead e.g. h1 -> h6

templates/home.html
```diff
@@ -1,17 +1,18 @@
...
-  <h2 class="greeting">Welcome {{ .data.email }}</h2>
-  <h2><a href="/keyword">Keyword</a></h2>
-  <form class="mt-8 space-y-6" action="/logout" method="POST">
-    <button type="submit">Logout</button>
-  </form>
-{{ else }}
-  <h2><a href="/login">Login</a></h2>
-  <h2><a href="/register">Register</a></h2>
-{{ end }}
+  <div class="home">
```
olivierobert: This class name should be added to the `body` tag. So each controller can correspond to one screen:

```
<body class="home">
<body class="keyword">
<body class="register">
```

This cannot be a component for sure. But also you often forget [a basic rule of components](https://nimblehq.co/compass/development/code-conventions/css/):

> Class names must be composed of at least two words to make each element class name specific and more descriptive.

olivierobert (changes requested):

assets/stylesheets/screens/home.scss
gutakk: Fixed in daa3a488ad94ff88a96a1ebc77bb2af1e75b3456

templates/home.html
gutakk: Fixed in a90e792bc33581d3d5c17bb0ad0bc1fd65f04217

gutakk: Fixed in 7a25f26094e351e41c3bf9355ca1d1fef2e16ee8

andyduong1920 (approved):

templates/home.html
gutakk: Fixed in 09a829ca2bb72bd502177a4df54f06df5b640a51

templates/home.html
```diff
@@ -1,17 +1,13 @@
...
-  <h2 class="greeting">Welcome {{ .data.email }}</h2>
-  <h2><a href="/keyword">Keyword</a></h2>
-  <form class="mt-8 space-y-6" action="/logout" method="POST">
-    <button type="submit">Logout</button>
-  </form>
-{{ else }}
-  <h2><a href="/login">Login</a></h2>
-  <h2><a href="/register">Register</a></h2>
-{{ end }}
+  <h4 class="display-subtitle sm:mt-5 md:text-xl">
```
olivierobert: These classes should be moved to the stylesheets instead (same issue above as well). Not sure why mixing inline Tailwind classes with `display-subtitle`.

olivierobert (changes requested):

templates/home.html
gutakk: Because Tailwind's `@apply` doesn't support non utility classes such as `media-query` or class that combine more than one css together such as `divide-y-0`.

But for `media-query`, I found a work around. We can do something like this
```css
h1 {
  @apply .text-size-3;
  @screen lg {
    @apply .text-size-5;
  }
}
```
Should we move the inline media-query class to the stylesheets as above example?

olivierobert: Yes, that's actually how we do it with Bootstrap too:=. Here is an example below from Compass's codebase:

https://github.com/nimblehq/compass/blob/83d644d3c6a91af2c0537216dbdab8f109a29b3e/_sass/screens/_team-developer-track.scss#L17-L27

templates/home.html
```diff
@@ -1,17 +1,13 @@
 {{ define "content" }}
+  {{ include "partials/flashes" }}
+  
+  <h1 class="sm:text-5xl md:text-6xl">
```
olivierobert: These styles must be added to the screen stylesheets if they don't fit into one component

templates/home.html
```diff
@@ -1,17 +1,13 @@
...
-  </form>
-{{ else }}
-  <h2><a href="/login">Login</a></h2>
-  <h2><a href="/register">Register</a></h2>
-{{ end }}
+  <h4 class="display-subtitle sm:mt-5 md:text-xl">
+    Gathering Google search information with your own keywords
+  </h4>
 
+  <a href="/keyword"><button class="button-primary sm:mt-8 sm:text-lg sm:px-8 sm:py-4">Get Started</button></a>
```
olivierobert: Same as above to move the styles to the screen sctylesheet

olivierobert (changes requested):

templates/home.html
hoangmirs: @gutakk yes, using `@screen` directive is a good solution. It would make the code clearer

gutakk: Fixed in a6e038f935ccaff5800ed8f2e6ddcfd998182677

templates/home.html
gutakk: Fixed in a6e038f935ccaff5800ed8f2e6ddcfd998182677

templates/home.html
gutakk: Fixed in a6e038f935ccaff5800ed8f2e6ddcfd998182677

olivierobert (approved):

PR #47, opened 2020-12-25
Setup OAuth2 JSON API

### LINKED ISSUE #9: Setup OAuth 2 Restful JSON API
### Acceptance Criteria
- Setup OAuth 2 Resful JSON API
- Include unit test

gutakk:
## What happened
Resolved #9 

## Insight
- [x] Setup OAuth2 server (store client and access token to postgres)
  - Create generate client API (return CLIENT_ID and CLIENT_SECRET) with basic auth
  - Create generate access token API
- [x] Add Dummy API to test token validation (client credentials grant type).

## Proof Of Work
- Generate client with valid basic auth
```http
# Request
curl --location --request POST 'localhost:8080/api/v1/client' --header 'Authorization: Basic YWRtaW46cGFzc3dvcmQ='

# Response
Status: 201 Created
{
    "data": {
        "id": "b995467a-aa62-43b3-9f58-e637fb53f032",
        "type": "client",
        "attributes": {
            "client_id": "b995467a-aa62-43b3-9f58-e637fb53f032",
            "client_secret": "5b946298-4fdf-4f49-aaee-7d4ba02e56d8"
        }
    }
}
```

- Generate client without basic auth or invalid basic auth
```http
# Request
curl --location --request POST 'localhost:8080/api/v1/client'

# Response
Status: 401 Unauthorized
```

- Generate token with valid client and secret (client credentials grant type)
``` http
# Request
curl --location --request POST 'localhost:8080/api/v1/token?grant_type=client_credentials&client_id=18281610-7448-4794-9e36-b87e1e205d81&client_secret=b052c23a-8f0a-4ffc-a4d5-f408f35e4de3'

# Response
Status: 200 OK
{
    "access_token": "OWEWNGFLZJUTMZVJMS0ZZTIWLWFHMJETOTM2ZJEWMTEWNDM1",
    "expires_in": 7200,
    "token_type": "Bearer"
}
```


.env.debug
```diff
@@ -1,3 +1,6 @@
 DB_NAME=go_google_scraper_development
 DB_PORT=5432
 REDIS_PORT=6379
+
+BASIC_AUTHENTICATION_USERNAME=admin
+BASIC_AUTHENTICATION_PASSWORD=password
```
olivierobert: Prefer keeping environment variables in alphabetical order and without line breaks to keep things compact here. Often these files must be created programmatically so line breaks are not something we want to handle 🙈 As for the alphabetical order, it allows to maintain those files in the long run as it usually keeps expanding.

controllers/api/oauth.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/google/uuid"
+)
+
+type OAuthController struct{}
+
+func (oa *OAuthController) ApplyRoutes(engine *gin.RouterGroup) {
+	engine.POST("/client", oa.generateClient)
+}
+
+func (oa *OAuthController) generateClient(c *gin.Context) {
```
olivierobert: That's business logic. Prefer to move it to a service module instead as done for the other controllers to keep the controllers "dumb" and lean. Even if it's used internally, having a separate package for it is better.

controllers/api/oauth_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/gutakk/go-google-scraper/tests/path_test"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/gin-gonic/gin"
+	"gopkg.in/go-playground/assert.v1"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+type OAuthClientResult struct {
```
olivierobert: Why not just simply `OAuthClient`?

main.go
```diff
@@ -17,6 +18,7 @@ func main() {
 	migration.Migrate(database)
 
 	db.SetupRedisPool()
+	oauth.SetupOAuthServer()
```
olivierobert: If any error occurs here, will this prevent the app to start? it should but since there is no error handling here, I am wondering what will happen 🤔

controllers/api/oauth_test.go
```diff
@@ -0,0 +1,103 @@
...
+	_ = json.Unmarshal(data, &respBody)
+
+	assert.Equal(s.T(), http.StatusCreated, resp.Code)
+	assert.NotEqual(s.T(), nil, respBody["CLIENT_ID"])
+	assert.NotEqual(s.T(), nil, respBody["CLIENT_SECRET"])
+
+	var result OAuthClientResult
+	db.GetDB().Table("oauth2_clients").Select("id", "secret", "domain").Scan(&result)
+
+	assert.Equal(s.T(), respBody["CLIENT_ID"], result.ID)
```
olivierobert: Having two sets of assertions is actually a code smell in tests 🙈 It also makes the tests harder to read. It's better to follow the basic rules to write steps in 23 stages: arrange (setup), execute and assert.

So:

```suggestion
headers := http.Header{}
// Basic auth with username = admin and password = password
headers.Set("Authorization", "Basic YWRtaW46cGFzc3dvcmQ=")

resp := testHttp.PerformRequest(s.engine, "POST", "/api/client", headers, nil)
data, _ := ioutil.ReadAll(resp.Body)
var respBody map[string]string
_ = json.Unmarshal(data, &respBody)
var result OAuthClientResult
db.GetDB().Table("oauth2_clients").Select("id", "secret", "domain").Scan(&result)

assert.Equal(s.T(), http.StatusCreated, resp.Code)
assert.NotEqual(s.T(), nil, respBody["CLIENT_ID"])
assert.NotEqual(s.T(), nil, respBody["CLIENT_SECRET"])
assert.Equal(s.T(), respBody["CLIENT_ID"], result.ID)
assert.Equal(s.T(), respBody["CLIENT_SECRET"], result.Secret)
assert.Equal(s.T(), fmt.Sprintf("http://localhost:%s", os.Getenv("APP_PORT")), result.Domain)
```

olivierobert (changes requested):
Do always add API versioning as best practice for the start when building REST API -> `api/v1`.

Overall it looks good and promising 💪

.env.debug
gutakk: Good catch. Fixed in bab4c3dc68c75bb97f09f317f4347526409344b9

controllers/api/oauth_test.go
gutakk: Fixed in c4e292105f2cd39a9f3f55e9f80cfedc5aa2ac31

gutakk:
> Do always add API versioning as best practice for the start when building REST API -> `api/v1`.
> 
> Overall it looks good and promising 💪

Should this be the new folder under the `api`? `api -> v1` or it can just be `api_v1`?

gutakk:
> > Do always add API versioning as best practice for the start when building REST API -> `api/v1`.
> > Overall it looks good and promising 💪
> 
> Should this be the new folder under the `api`? `api -> v1` or it can just be `api_v1`?

Fixed the first versioning in c07d662fabec00ea6ca93a738bfb6d995ef0a40a

controllers/api/oauth_test.go
gutakk: `OAuthClient` looks more meaningful. Fixed in f593909dd97a534fa2091abec4e6c3c0ff275a68

main.go
gutakk: My mistake. I read their document and they don't handling the error.
Handling error in bec83f3ffa6dd47361c0d2e2c247b86d61eafe08

controllers/api/oauth.go
gutakk: Separated in 7215fcf81efa4af24e664148e28044bd9ce13d8e

gutakk:
@olivierobert
Updated response to follow JSON API convention in 2443f4a65cd42c8abe479bbda5901ae57a01ef43 148adcc425ef0ed1d5ad58fc7f36b1073b7729a3

andyduong1920 (approved):

controllers/api_v1/dummy.go
```diff
@@ -0,0 +1,19 @@
+package api_v1
```
andyduong1920: Should we remove this controller as we don't need that on production @gutakk, this is something useful on `development` phase... as this PR gonna merge we should remove it.

Another thing, we don't have `dummy_test` so not sure what's the goal to keep it

gutakk: Removed in this [PR](https://github.com/gutakk/go-google-scraper/pull/53)

andyduong1920: I saw that but why don't we remove in this PR as this PR gonna merge to develop branch before that PR :D

gutakk: That make sense. Removed in bee4297ad3e629a4ba78b8a34cb05d08386b3eda

services/oauth_service/client.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func GenerateClient() (OAuthClient, error) {
+	clientID := uuid.New().String()
+	clientSecret := uuid.New().String()
+
+	err := oauth.GetClientStore().Create(&models.Client{
+		ID:     clientID,
+		Secret: clientSecret,
+		Domain: fmt.Sprintf("http://localhost:%s", os.Getenv("PORT")),
```
andyduong1920: should we move the `localhost` into ENV 🤔

gutakk: Good catch. Updated in 11a0f2f9a7377090fdd89b756b6f3b79dbf12d0c

helpers/api_helper/json_api_response.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "github.com/gin-gonic/gin"
+
+type ErrorResponseObject struct {
+	Title  string `json:"title"`
+	Detail string `json:"detail"`
+	Status int    `json:"status"`
+}
+
+func (e *ErrorResponseObject) ConstructErrorResponse() gin.H {
```
olivierobert: `Construct` is not very common naming, how about to use a more generic term like `BuildErrorResponse` (there is also such a thing called [the Builder pattern](https://golangbyexample.com/builder-pattern-golang/)). But maybe here since it's returning a JSON based on the attributes params, how about just:

```suggestion
func (e *ErrorResponseObject) NewErrorResponse() gin.H {
```

Then in the controller, it twould be:

```go
c.JSON(errorResponse.Status, errorResponse.NewErrorResponse())
```

main.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/gutakk/go-google-scraper/controllers"
 	"github.com/gutakk/go-google-scraper/db"
 	"github.com/gutakk/go-google-scraper/migration"
+	"github.com/gutakk/go-google-scraper/oauth"
+)
+
+const (
+	startServerFail = "Failed to start the server %v"
```
olivierobert: Overall, be more specific in the naming of constants as I am often a bit lost of the usage of each constant. Some are to define view name while other error messages. So how about using a suffix to be more clear:

```suggestion
	startServerFailureError = "Failed to start the server %v"
```
or

```suggestion
	startServerFailureMessage = "Failed to start the server %v"
```

controllers/api_v1/token.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+type TokenAPIController struct{}
+
+// TODO: Unit test in login API PR as grant type need to change to password
+func (t *TokenAPIController) ApplyRoutes(engine *gin.RouterGroup) {
+	engine.POST("/token", t.generateToken)
+}
+
+func (t *TokenAPIController) generateToken(c *gin.Context) {
```
olivierobert: While creating a token is a small piece of code and so seems it could be added to the controller, but this is not consistent with `OAuthController` which uses the OAuth Service package. Also, this controller now has a hard dependency on the package `oauth`. As discussed before, only exposes services to the controllers for better architecture patterns. 
So I recommend moving the `generateToken` method to the OAuth Service package.

olivierobert (changes requested):
Looks good overall 💪

helpers/api_helper/json_api_response.go
gutakk: Make sense. Fixed in 23688b8db37ebc32331c1a92c71308cdd78010d9

main.go
gutakk: Good point. Fixed in 8b3ce7a0c3a1f4848846110992dce4d2f1748a9b

controllers/api_v1/token.go
gutakk: Fixed in b09ee8387fbed21b7a76ec97738e0d11915e3794

olivierobert (approved):

PR #56, opened 2021-01-13
As a user, I can filter my keywords by name

### LINKED ISSUE #10: As a user, I can filter my keywords by name
### Acceptance Criteria
- Implement filter by name feature
- Include unit test

gutakk:
## What happened
Resolve #10 

## Insight
- [x] Update controller to support filter by query strings
- [x] Add keyword filter section

## Proof Of Work
- Search by keyword title (found)
Screen Shot 2564-01-14 at 14 32 39
Screen Shot 2564-01-14 at 14 41 54

- Not found
Screen Shot 2564-01-14 at 15 10 08


services/keyword_service/keyword.go
```diff
@@ -28,11 +30,10 @@ type KeywordService struct {
 	CurrentUserID uint
 }
 
-func (k *KeywordService) GetAll() ([]models.Keyword, error) {
-	condition := make(map[string]interface{})
-	condition["user_id"] = k.CurrentUserID
+func (k *KeywordService) GetAll(conditions []string) ([]models.Keyword, error) {
```
olivierobert: `GetAll` is kind of misleading since it will be a filtered list of keywords. How about following the same pattern as the models and naming it `GetKeywords`?

services/keyword_service/filter.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	keywordTitleQuery = "keyword-title"
+	keywordTitleDBCol = "keyword"
+)
+
+func GetKeywordConditionsFromQueryStrings(queryStrings map[string][]string) []string {
+	var conditions []string
+
+	title := ensureQueryKeyExist(queryStrings, keywordTitleQuery)
+	if title != "" {
```
olivierobert: The naming is also confusing here as there is no such thing as a keyword title 😵 Naming is hard but this creates lots of confusion

templates/keyword.html
```diff
@@ -13,6 +13,13 @@ <h2>Upload CSV</h2>
     </button>
   </form>
 
+  <form class="form-keyword-filter space-y-6">
```
olivierobert: This form is missing its action and method attributes. When omitted it's submitted to the current page as `GET` so it works but still, it's not a valid form.

templates/partials/filter_keyword_input.html
```diff
@@ -0,0 +1,12 @@
+{{ define "filter_keyword_input" }}
+
+<div class="space-y-2">
+  <label>{{ .filterLabel }}</label>
+  {{ $queryValue := "" }}
+  {{ if .query }}
+    {{ $queryValue = index .query .filterName 0 }}
```
olivierobert: Why is this value computed in the view? This hides lots of logic. I recommend that the controller transforms the value so the view can just render it. When the logic is too complex, that's what presenters classes are for.

controllers/keyword.go
```diff
@@ -136,6 +137,7 @@ func getKeywordsData(keywordService keyword_service.KeywordService) map[string]i
 
 	data := getCurrentUser(keywordService)
 	data["keywordPresenters"] = keywordPresenters
+	data["query"] = queryString
```
olivierobert: If we consider that filtering is performing a search, then I have a preference to name this:

```suggestion
	data["search"] = queryString
```

of 

```suggestion
	data["filter"] = queryString
```

`queryString` feels inspired by the fact that the search value is one of the query string items 😁  So instead of naming it after where it comes from, name it after what it does. Say the form is changed to `POST` then this variable would still mean the same.

controllers/keyword_test.go
```diff
@@ -84,6 +84,23 @@ func TestDisplayKeywordWithAuthenticatedUser(t *testing.T) {
...
+func TestDisplayKeywordWithAuthenticatedUserWithFilter(t *testing.T) {
+	engine := testConfig.GetRouter(true)
+	new(KeywordController).applyRoutes(EnsureAuthenticatedUserGroup(engine))
+
+	// Cookie from login API Set-Cookie header
+	headers := http.Header{}
+	cookie := fixture.GenerateCookie("user_id", "test-user")
+	headers.Set("Cookie", cookie.Name+"="+cookie.Value)
+
+	response := testHttp.PerformRequest(engine, "GET", "/keyword?keyword-title=Test", headers, nil)
```
olivierobert: Kind of related to the comment above, the following query string would make more sense:

```suggestion
	response := testHttp.PerformRequest(engine, "GET", "/keyword?search=test", headers, nil)
```

controllers/keyword_test.go
```diff
@@ -84,6 +84,23 @@ func TestDisplayKeywordWithAuthenticatedUser(t *testing.T) {
...
+	p, err := ioutil.ReadAll(response.Body)
+	isKeywordPage := err == nil && strings.Index(string(p), "<title>Keyword</title>") > 0
```
olivierobert, line 98: This check is kind of weak (it's actually used in many tests) as it does not tell us much specificity about this test itself. Moving forward this thinking, just like on other platforms there should be two kinds of unit tests:
- Controllers tests: check that the correct view is rendered with the proper variables
- Requests tests: check for HTTP status codes

Now, in this project, the controller tests do well the HTTP status code checks but the checks for the view rely on HTML code checks. 

I don't expect changes in this PR. This is meant to guide you in the right direction for improving the codebase.

helpers/view/filter_keyword_input.go
```diff
@@ -0,0 +1,10 @@
+package view
+
+func FilterKeywordPartialInput(query map[string][]string, label string, name string, placeholder string) map[string]interface{} {
```
olivierobert: Not sure why `query` is defined as a slice of string 🤔 Is there any expectation to have more than one search term?

helpers/view/filter_keyword_input_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"gopkg.in/go-playground/assert.v1"
+)
+
+func TestFilterKeywordPartialInputWithValidParams(t *testing.T) {
+	query := map[string][]string{
+		"testQuery": {"Lorem Ipsum"},
+	}
+	result := view.FilterKeywordPartialInput(query, "testLabel", "testName", "testPlaceholder")
+
+	assert.Equal(t, map[string][]string{"testQuery": {"Lorem Ipsum"}}, result["query"])
```
olivierobert: Related to above but woudln't the following be enough:

```suggestion
	assert.Equal(t, "Lorem Ipsum", result["query"])
```

models/keyword.go
```diff
@@ -56,7 +56,10 @@ func GetKeywordBy(condition map[string]interface{}) (Keyword, error) {
 	return keyword, nil
 }
 
-func GetKeywordsBy(condition map[string]interface{}) ([]Keyword, error) {
+func GetKeywordsBy(condition interface{}) ([]Keyword, error) {
```
olivierobert: Why allow any kind of data for `condition`?  Wouldn't it better to be more specific and to allow multiple conditions:

```suggestion
func GetKeywordsBy(condition []map[string]string) ([]Keyword, error) {
```

models/keyword_test.go
```diff
@@ -203,7 +204,20 @@ func (s *KeywordDBTestSuite) TestGetKeywordsByValidKeyword() {
...
 }
 
-func (s *KeywordDBTestSuite) TestGetKeywordsByInvalidKeyword() {
+func (s *KeywordDBTestSuite) TestGetKeywordsByValidKeywordStringCondition() {
+	keyword := Keyword{UserID: s.userID, Keyword: faker.Name()}
+	db.GetDB().Create(&keyword)
+
+	condition := fmt.Sprintf("keyword = '%s'", keyword.Keyword)
+
+	result, err := GetKeywordsBy(condition)
```
olivierobert: `GetKeywordsBy` should receive a map or slice of maps instead of readily formatted string as it pushes the responsibility to the consumer of the method to send `condition` as "keyword = '%s'". 

Related to the above comment, this is an method signature that would be more standard:

```suggestion
	result, err := GetKeywordsBy([{"keyword": "test"}])
```

services/keyword_service/filter.go
```diff
@@ -0,0 +1,29 @@
...
+		conditions = append(conditions, fmt.Sprintf("LOWER(%s) LIKE LOWER('%%%s%%')", keywordTitleDBCol, title))
```
olivierobert: The SQL formatting `LIKE LOWER` should be done in the model method. The service should just downcase if needed but not adding `%` to the query string.

services/keyword_service/filter.go
```diff
@@ -0,0 +1,29 @@
+package keyword_service
+
+import (
+	"fmt"
+)
+
+const (
+	keywordTitleQuery = "keyword-title"
```
olivierobert: This couples very tightly the service with the current HTML view. It should not be the case. Just think for the API, this will not be needed.  

In the above comments, I am recommending to simplify a lot the implementation. Checking for presence of the query parameters should not require such complex logic. If an empty value is sent, just validate it on the controller just like a regular missing value. 

So I recommend removing entirely this file.

services/keyword_service/keyword.go
```diff
@@ -28,11 +30,10 @@ type KeywordService struct {
...
+	conditions = append(conditions, fmt.Sprintf("user_id = '%s'", fmt.Sprint(k.CurrentUserID)))
 
-	keywords, err := models.GetKeywordsBy(condition)
+	keywords, err := models.GetKeywordsBy(strings.Join(conditions, " AND "))
```
olivierobert: The concatenation of the condition should be done in the model class. `GetKeywordsBy` should receive a slice and create the SQL query, not the services. This would create duplication of logic if `GetKeywordsBy` must be used in several services.

olivierobert (changes requested):
There are major issues in where the business logic and data transformation are happening. So let's recap what should be the adequate way:
- View send the search/filter keyword via  GET e.g. `?search=test`
- The controller returns an error if `search=` is nil otherwise send the search/filter keyword to the service as a map `{"search": "test"}` or a slice of maps
- The service receives the search param and send it to the model
- The model generates the SQL query to filter the keywords based on a map of conditions

controllers/keyword_test.go
gutakk: How about if we have more than one searching params as I split them into this 2 PRs
https://github.com/gutakk/go-google-scraper/pull/61 https://github.com/gutakk/go-google-scraper/pull/62

helpers/view/filter_keyword_input.go
gutakk: Because `c.Request.URL.Query()` return map[string][]string and from this [comment](https://github.com/gutakk/go-google-scraper/pull/56#discussion_r562531927), yes there will be more than one search term.

olivierobert: I understand now :-)

controllers/keyword_test.go
olivierobert: So, if there are multiple search terms, it should be an array:

```
"/keyword?search[]=test&search[]=something"
```

If there are different types of search params, it should be an associative array:

```
"/keyword?search['keyword']=test&search['is_adword_advertiser']=true"
```

services/keyword_service/keyword.go
gutakk: Fixed in de06907c10c1d49f3a385efd869d4bff8dd58d9c

templates/keyword.html
gutakk: Added action and method in a3d34048bc51c14100cf7de864c7c822ed54e886

controllers/keyword.go
gutakk: That's make sense. Fixed in 684f91a8e9a45e90450d6bc9a41967df0ca60a8c

controllers/keyword_test.go
gutakk: Fixed in 40ffebae788fcd1829b0016679ccc00d311d8528

controllers/keyword_test.go
gutakk: Will fix this in this https://github.com/gutakk/go-google-scraper/issues/69

helpers/view/filter_keyword_input_test.go
gutakk: Fixed in 40c1966afcfd1b1a68cdeedbcfb92c67cc7e8b02

templates/partials/filter_keyword_input.html
gutakk: Fixed in 40c1966afcfd1b1a68cdeedbcfb92c67cc7e8b02

models/keyword.go
gutakk: Because gorm `Where` function allow both map and string condition. You can pass both following to the function
```go
db.GetDB().Where("user_id = 1").First(&keyword)
```
```go
condition = make(map[string]interface{})
condition["user_id"] = 1
db.GetDB().Where(condition).First(&keyword)
```
But I think your suggestion is better as we can force the type of data to pass in and it is easier to handle.
Fixed in 65afd1cc44daba32fc22b257bfc5bb964b76564b

models/keyword_test.go
gutakk: Fixed in 65afd1cc44daba32fc22b257bfc5bb964b76564b

services/keyword_service/keyword.go
gutakk: Fixed in https://github.com/gutakk/go-google-scraper/pull/56#issuecomment-766715632

services/keyword_service/filter.go
gutakk: Fixed in https://github.com/gutakk/go-google-scraper/pull/56#issuecomment-766715632

gutakk:
> There are major issues in where the business logic and data transformation are happening. So let's recap what should be the adequate way:
> 
> * View send the search/filter keyword via  GET e.g. `?search=test`
> * The controller returns an error if `search=` is nil otherwise send the search/filter keyword to the service as a map `{"search": "test"}` or a slice of maps
> * The service receives the search param and send it to the model
> * The model generates the SQL query to filter the keywords based on a map of conditions

I followed these steps in 3f8e39ac0f70d07806a122e48e9ae6466b5bf615 ad8354900b1e9ef2743ef49b0e39bfd5e275ec09.
And I think that query string and db mapper should be done in model, so I decided to move it in b0dfda16a2ada86563ab9d0ad5dbc539b879d353

**Update Jan 26, 2021**
Decided to use string statement instead of generic where type in 5a53d35cc9cd6b89754f6f94e07d1e9f2e5f95bf 1757094882e2c67709d67f10c452ab3f5e699b17

services/keyword_service/filter.go
gutakk: Removed filter in a1f195c85d356b6a110e52584f5aec3db829854a

services/keyword_service/filter.go
gutakk: Removed filter in a1f195c85d356b6a110e52584f5aec3db829854a

models/keyword.go
```diff
@@ -18,9 +20,21 @@ const (
 	Processed  KeywordStatus = "processed"
 	Failed     KeywordStatus = "failed"
 
-	InvalidKeywordStatusErr = "invalid keyword status"
+	InvalidKeywordStatusErr    = "invalid keyword status"
+	couldNotJoinConditionError = "could not join conditions"
 )
 
+type Condition struct {
```
olivierobert: As an improvement, since this struct could be used in other models, this struct and the method `getJoinedConditions` could be moved to another file in `models` package

controllers/keyword.go
```diff
@@ -22,6 +24,13 @@ const (
 	uploadSuccessFlash = "CSV uploaded successfully"
 )
 
+var FilterList = []map[string]string{
+	{
+		"queryString": "filter[keyword]",
+		"filter":      "keyword",
```
olivierobert: At first I thought that there were two filters but I was wrong since there is only only `filter[keyword`. The second item is used to generate the model condition. So It's a bit hard to understand that from the get-go.

How about renaming the second key to `modelCondition` so that it's more explicit:

```suggestion
		"modelCondition":      "keyword",
```

olivierobert (commented):
Great improvement. So minor comments 👍

controllers/keyword.go
gutakk: That is a good idea. Fixed in d2064d29cb90fca0dfc74b75bf9ced442e04278a

models/keyword.go
gutakk: Good catch. Fixed in a1d1a1d1ffd6fb5cb7ad53c944104ca1c82e97d9

assets/stylesheets/screens/_keyword.scss
```diff
@@ -1,9 +1,15 @@
 body.keyword {
-  .form-keyword-upload {
+  .form-keyword-upload, .form-keyword-filter {
```
carryall: @gutakk Can we move `.form-keyword-filter` to another line? 🙌🏼 
```suggestion
  .form-keyword-upload,
  .form-keyword-filter {
```

controllers/keyword.go
```diff
@@ -136,10 +146,27 @@ func getKeywordsData(keywordService keyword_service.KeywordService) map[string]i
...
+	data["filter"] = queryString
 
 	return data
 }
 
+func validateValidConditions(queryString map[string][]string) []models.Condition {
+	var validConditions []models.Condition
+
+	for _, f := range FilterList {
+		queryStringValue := queryString[f["queryString"]]
```
carryall, line 158: @gutakk Will `queryString` be nil when we call `getKeywordsData(keywordService, nil)`? 🤔

assets/stylesheets/screens/_keyword.scss
gutakk: Absolutely, fixed in 74316d6dcaad1f320cd8f82712cfe68c1a271ae2

controllers/keyword.go
gutakk: Exactly, that's mean no filter at all.

controllers/keyword.go
```diff
@@ -136,10 +146,27 @@ func getKeywordsData(keywordService keyword_service.KeywordService) map[string]i
 
 	data := getCurrentUser(keywordService)
 	data["keywordPresenters"] = keywordPresenters
+	data["filter"] = queryString
 
 	return data
 }
 
+func validateValidConditions(queryString map[string][]string) []models.Condition {
```
carryall: @gutakk How about changing the function name to `filterValidConditions` or `validConditions` so it's more aligned with the function behavior, IMO naming it `validate` sound like it would return boolean or add some error if the input does not pass the validation 🤔

olivierobert: Indeed it could be a good naming improvement. Just like in Rails there are strong params to permit only certain params.

olivierobert (approved):
It's a massive improvement from the first iteration 💪 There are a few naming improvements that could be added in a few places but overall this looks good to me

controllers/keyword.go
gutakk: Good point! Fixed in 2060a940e9d85e9b31d4d90a3e92a4539b2b4268

carryall (approved):

assets/stylesheets/screens/_keyword.scss
```diff
@@ -1,9 +1,16 @@
 body.keyword {
-  .form-keyword-upload {
+  .form-keyword-upload,
+  .form-keyword-filter {
     @apply border rounded-md p-6;
     width: fit-content;
```
hoangmirs, line 5: Not related to this PR, but you can customize the width in `tailwind.config.js file`, then you can re-use it
https://v1.tailwindcss.com/docs/width#customizing
```
// tailwind.config.js
module.exports = {
    theme: {
      extend: {
        width: {
           "fit-content": "fit-content";
        }
      }
    }
  }

// Then you can use:
@apply w-fit-content;
```

hoangmirs (approved):

assets/stylesheets/screens/_keyword.scss
gutakk: That looks great. So I can add any style I want into it, right? such as sometimes they don't provide `w-11`.

PR #61, opened 2021-01-15
As a user, I can filter keyword by URL in the result

### LINKED ISSUE #58: As a user, I can filter keyword by URL in the result
## Acceptance Criteria
- Implement filter by URL in the result feature
- Include unit test

gutakk:
## What happened
Resolve #58 

## Insight
- [x] Add input field for URL filtering
- [x] Add URL filtering logic

## Proof Of Work
- Filter using URL
Screen Shot 2564-01-15 at 12 29 15

- Filter combine title and URL
Screen Shot 2564-01-15 at 12 29 46

- Not found
Screen Shot 2564-01-15 at 12 30 09

### DIFF
--- a/controllers/keyword.go
+++ b/controllers/keyword.go
@@ -29,6 +29,10 @@ var FilterList = []map[string]string{
 		"queryString":    "filter[keyword]",
 		"modelCondition": "keyword",
 	},
+	{
+		"queryString":    "filter[url]",
+		"modelCondition": "url",
+	},
 }
 
 type KeywordController struct{}
--- a/models/keyword_condition.go
+++ b/models/keyword_condition.go
@@ -18,6 +18,7 @@ type Condition struct {
 // Map query string filter to where statement
 var ConditionMapper = map[string]string{
 	"keyword": "LOWER(keyword) LIKE LOWER('%%%s%%')",
+	"url":     "(LOWER(non_adword_links::text) LIKE '%%%[1]s%%' OR LOWER(top_position_adword_links::text) LIKE '%%%[1]s%%')",
 	"user_id": "user_id = '%s'",
 }
 
--- a/assets/stylesheets/screens/_keyword.scss
+++ b/assets/stylesheets/screens/_keyword.scss
@@ -9,6 +9,10 @@ body.keyword {
     &__input-container {
       @apply flex;
     }
+
+    &__button {
+      @apply w-24;
+    }
   }
   
   .list-keyword {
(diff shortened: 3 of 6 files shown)


olivierobert (approved):

controllers/keyword.go
```diff
@@ -29,6 +29,10 @@ var FilterList = []map[string]string{
 		"queryString":    "filter[keyword]",
 		"modelCondition": "keyword",
 	},
+	{
+		"queryString":    "filter[url]",
+		"modelCondition": "url",
+	},
```
olivierobert, line 35: With the refactor, it's now easier to add new filters 💪

models/keyword_condition.go
```diff
@@ -18,6 +18,7 @@ type Condition struct {
 // Map query string filter to where statement
 var ConditionMapper = map[string]string{
 	"keyword": "LOWER(keyword) LIKE LOWER('%%%s%%')",
+	"url":     "(LOWER(non_adword_links::text) LIKE '%%%[1]s%%' OR LOWER(top_position_adword_links::text) LIKE '%%%[1]s%%')",
```
carryall, line 21: @gutakk One small question, does using `LOWER()` instead of `ILIKE` here make it better in terms of performance?

carryall (approved):

models/keyword_condition.go
gutakk: This is the result using `EXPLAIN ANALYZE` for 60 rows data with following query
- Using LOWER LIKE
```sql
EXPLAIN ANALYZE SELECT * FROM keywords WHERE (LOWER(non_adword_links::text) LIKE '%wiki%' OR LOWER(top_position_adword_links::text) LIKE '%wiki%');

Seq Scan on keywords  (cost=0.00..7.80 rows=1 width=657) (actual time=0.186..1.391 rows=21 loops=1)
   Filter: ((lower((non_adword_links)::text) ~~ '%wiki%'::text) OR (lower((top_position_adword_links)::text) ~~ '%wiki%'::text))
   Rows Removed by Filter: 39
Planning Time: 0.975 ms
Execution Time: 1.489 ms
(5 rows)
```

- Using ILIKE
```sql
EXPLAIN ANALYZE SELECT * FROM keywords WHERE non_adword_links::text ILIKE '%wiki%' OR top_position_adword_links::text ILIKE '%wiki%';

Seq Scan on keywords  (cost=0.00..7.50 rows=1 width=657) (actual time=0.081..2.509 rows=21 loops=1)
   Filter: (((non_adword_links)::text ~~* '%wiki%'::text) OR ((top_position_adword_links)::text ~~* '%wiki%'::text))
   Rows Removed by Filter: 39
Planning Time: 0.262 ms
Execution Time: 2.550 ms
(5 rows)
```

And here is some more performance discussion
https://stackoverflow.com/questions/20336665/lower-like-vs-ilike

PR #62, opened 2021-01-15
As a user, I can filter keyword by Adwords existence

### LINKED ISSUE #59: As a user, I can filter keyword by Adwords existence
## Acceptance Criteria
- Implement filter by Adwords existence feature
- Include unit test

gutakk:
## What happened
Resolve #59 

## Insight
- [x] Add input field (checkbox) for Adwords existence filtering
- [x] Add Adwords existence filtering logic

## Proof Of Work
- Filter using adwords
Screen Shot 2564-02-02 at 11 07 57


assets/stylesheets/screens/_keyword.scss
```diff
@@ -13,6 +13,12 @@ body.keyword {
     &__button {
       @apply w-24;
     }
+
+    &__checkbox-input {
+      @apply block;
+      zoom: 3;
```
olivierobert: `zoom` is [a non-standard CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom) so should be avoided.

In terms of UI, I am pretty sure you never saw a checkbox that big in your life 😆 While it's not a UX/UI contest, do try to think of standard patterns. In this case, having the checkbox on a different row than the inputs would be better:

image

controllers/keyword.go
```diff
@@ -26,12 +27,16 @@ const (
 
 var FilterList = []map[string]string{
 	{
-		"queryString":    "filter[keyword]",
-		"modelCondition": "keyword",
+		"queryString":    fmt.Sprintf("filter[%s]", models.KeywordCondition),
```
olivierobert: In some cases, DRY is not ideal. This new implementation couples again the controller with the model which is not great in terms of architecture. 

The frontend form input could be different but not it looks like it must be the same; that's not good flexibility. The following could be fine:
```
{
  "queryString":    "filter[text]",
  "modelCondition": models.KeywordCondition,
}
```

So prefer using the constant only for the `modelCondition` key.

On another perspective, the new implementation also lacks explicitness and clarity. Prefer simpler solutions 😌

models/keyword_condition.go
```diff
@@ -31,7 +38,12 @@ func GetJoinedConditions(conditions []Condition) (string, error) {
 		whereStatement := ConditionMapper[conditionName]
 
 		if conditionValue != "" && whereStatement != "" {
-			formattedConditions = append(formattedConditions, fmt.Sprintf(whereStatement, conditionValue))
+			switch conditionName {
```
olivierobert: Don't use case/switch blocks when there are only two cases if/else is best for that

olivierobert (changes requested):

models/keyword_condition.go
gutakk: Fixed in f4bd4c599937801d324fa45c0a54475b882e1bac

controllers/keyword.go
gutakk: Good catch. Fixed in dedc8c3b8cf24cfdfbb56e1d83a1905437990e2b

assets/stylesheets/screens/_keyword.scss
gutakk: I agree, I never see checkbox like I did before 😂 Fixed in 70764dd1ef8b2b090a7c4aed0bea42f032daa628

olivierobert: @gutakk Please update the PR description with the updated UI 🙏

gutakk: Updated 😅

olivierobert (approved):

carryall (approved):

Issue #36, opened 2020-12-17
Improve the test for Gin


gutakk:
focus in proposing some changes for one type of tests (models or controllers). Then we agree on the changes and you make all changes in multiple PRs.

PR #75, opened 2021-02-01
Fix cannot redirect to register page

### LINKED ISSUE #74: Cannot redirect to register page
## Issue
Cannot redirect to login/register page when no user in database but cookie exists in browser.

## Expected
Can redirect to login/register page even cookie exists in browser but no user in database.

gutakk:
### Issue
@carryall found that she cannot access to register page, app keep redirect to the root path.
I found that because I check only for if userID cookie exist in the browser but not in the database then user cannot access to auth feature and able to access to authenticated app feature. So I updated middleware to check if userID cookie also exist in the database.

Resolve #74 

## Insight
- [x] Update ensure guest and authenticated user middleware to check user id in the cookie with database
- [x] Add test cases for this bug

## Proof Of Work
Fixed the problem described in `What Happened`

### DIFF
--- a/controllers/logout_test.go
+++ b/controllers/logout_test.go
@@ -1,14 +1,19 @@
 package controllers
 
 import (
+	"log"
 	"net/http"
 	"testing"
 
-	"github.com/gin-gonic/gin"
+	"github.com/gutakk/go-google-scraper/db"
+	"github.com/gutakk/go-google-scraper/models"
 	testConfig "github.com/gutakk/go-google-scraper/tests/config"
 	"github.com/gutakk/go-google-scraper/tests/fixture"
 	testHttp "github.com/gutakk/go-google-scraper/tests/http"
+
+	"github.com/gin-gonic/gin"
 	"github.com/stretchr/testify/suite"
+	"golang.org/x/crypto/bcrypt"
 	"gopkg.in/go-playground/assert.v1"
 )
 
@@ -27,7 +32,14 @@ func TestLogoutTestSuit(t *testing.T) {
 }
 
 func (s *LogoutTestSuite) TestLogoutWithAuthenticatedUser() {
-	cookie := fixture.GenerateCookie("user_id", "test-user")
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte("testPassword"), bcrypt.DefaultCost)
+	if err != nil {
+		log.Fatalf("Cannot hash password: %s", err)
+	}
+	user := models.User{Email: "test@email.com", Password: string(hashedPassword)}
+	db.GetDB().Create(&user)
+
+	cookie := fixture.GenerateCookie("user_id", user.ID)
 	headers := http.Header{}
 	headers.Set("Cookie", cookie.Name+"="+cookie.Value)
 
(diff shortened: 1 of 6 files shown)


middlewares/ensure_guest_user.go
```diff
@@ -3,15 +3,25 @@ package middlewares
...
 func EnsureGuestUser(c *gin.Context) {
 	userID := session.Get(c, "user_id")
 
 	if userID != nil {
-		c.Redirect(http.StatusFound, "/")
+		_, err := models.FindUserByID(userID)
+
+		if err != nil {
+			session.Delete(c, "user_id")
+			c.Redirect(http.StatusFound, c.FullPath())
```
olivierobert: What would be `c.FullPath()` 🤔 I am not sure if there is a need for different redirects

gutakk: Good catch. I did it in the complex way.
Fixed in 4e70d7f38f4c8664a2d3c60fb73f7474eebefd36 9bd6b1b9ae9f1ee19d3a92b1bdfdc913bd8066ec

carryall (approved):

controllers/logout_test.go
```diff
@@ -27,7 +32,14 @@ func TestLogoutTestSuit(t *testing.T) {
 }
 
 func (s *LogoutTestSuite) TestLogoutWithAuthenticatedUser() {
-	cookie := fixture.GenerateCookie("user_id", "test-user")
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte("testPassword"), bcrypt.DefaultCost)
+	if err != nil {
+		log.Fatalf("Cannot hash password: %s", err)
+	}
+	user := models.User{Email: "test@email.com", Password: string(hashedPassword)}
+	db.GetDB().Create(&user)
```
carryall, line 40: @gutakk How about moving this to a helper function? I think we can put it in `fixture` 🤔

middlewares/ensure_authenticated_user.go
```diff
@@ -3,12 +3,13 @@ package middlewares
...
 
-	"github.com/gin-gonic/gin"
 	session "github.com/gutakk/go-google-scraper/helpers/session"
+
+	"github.com/gin-gonic/gin"
 )
 
 func EnsureAuthenticatedUser(c *gin.Context) {
-	userID := session.Get(c, "user_id")
+	userID := c.MustGet("currentUser")
```
olivierobert, line 12: What is the benefit of using `MustGet` over `session.Get`?

controllers/logout_test.go
gutakk: I will refactor this in the refactor test [PR](https://github.com/gutakk/go-google-scraper/issues/36) 🙏

middlewares/ensure_authenticated_user.go
gutakk: `session.Get` will get the cookie from browser directly, which I had it in the `current_user` middleware. `current_user` middleware is the first step of the route validation and also had a logic to check for user in the database with `user_id` and store the `currentUser` to gin context.

`MustGet` is the function to get value from gin context from specific key.

olivierobert (approved):

PR #78, opened 2021-02-08
Refactor code to follow convention

### LINKED ISSUE #33: Refactor code to follow convention in compass

gutakk:
## What happened
Resolve #33 

## Insight
- Refactor code in development branch to follow Go convention in compass development branch.

**Note**
This branch is only refactor to align with our Go convention. I create https://github.com/gutakk/go-google-scraper/issues/79 to move duplicate logic in unit test to helper function.

## Proof Of Work
- All tests must be passed.
- App still working properly.


helpers/error_handler/error_constant.go
```diff
@@ -0,0 +1,19 @@
...
+	JSONMarshalFailure       = "Failed to marshal json: "
+	JSONUnmarshalFailure     = "Failed to unmarshal json: "
+	MigrateDatabaseFailure   = "Failed to migrate database: "
+	ReadResponseBodyFailure  = "Failed to read response body: "
+	RecordInitializeFailure  = "Failed to init record: "
+	RecordStopFailure        = "Failed to stop record: "
+	RequestInitializeFailure = "Failed to init request: "
+	RequestFailure           = "Failed to make a request: "
+	ScanRowFailure           = "Failed to scan row: "
+	StartOAuthServerFailute  = "Failed to setup oauth server: "
```
carryall:
```suggestion
	StartOAuthServerFailure  = "Failed to setup oauth server: "
```

helpers/session/session.go
```diff
@@ -3,22 +3,31 @@ package helpers
 import (
 	"github.com/gin-contrib/sessions"
 	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
 )
 
 const (
 	sessionTimeoutInSec = 60 * 60 * 24 // 24 hours
+
+	saveSessionFailure = "Failed to save session: "
```
carryall, line 13: @gutakk  Should we move this to `helpers/error_handler/error_constant.go`? 🤔

services/google_search_service/request_test.go
```diff
@@ -5,48 +5,69 @@ import (
...
+	if err != nil {
+		log.Error(errorHelper.RecordStopFailure, err)
+	}
 
 	p, err := ioutil.ReadAll(resp.Body)
 	isGoogleSearchPage := err == nil && strings.Index(string(p), "<title>AWS") > 0
 
-	assert.Equal(t, nil, err)
+	assert.Equal(t, nil, requestErr)
 	assert.Equal(t, true, isGoogleSearchPage)
```
carryall, line 32: @gutakk Do you think it would be easier to read if we handle the error from `ioutil.ReadAll` separately before assertion?
```Go
p, err := ioutil.ReadAll(resp.Body)
if err != nil {
  log.Error(...)
}

isGoogleSearchPage := strings.Index(string(p), "<title>AWS") > 0
assert.Equal(t, true, isGoogleSearchPage)
```

services/keyword_service/keyword.go
```diff
@@ -62,31 +63,31 @@ func (k *KeywordService) Save(parsedKeywordList []string) error {
...
 
-			enqueueErr := google_search_service.EnqueueSearchJob(savedKeyword)
-			if enqueueErr != nil {
-				return enqueueErr
+			err = google_search_service.EnqueueSearchJob(savedKeyword)
+			if err != nil {
+				return err
 			}
 
 			return nil
```
carryall: @gutakk How about this?
```
return google_search_service.EnqueueSearchJob(savedKeyword)
```

services/keyword_service/keyword.go
```diff
@@ -107,9 +108,17 @@ func (k *KeywordService) ReadFile(filename string) ([]string, error) {
 
 func (k *KeywordService) UploadFile(c *gin.Context, file *multipart.FileHeader) string {
 	path := "dist/"
-	_ = os.Mkdir(path, 0755)
+	err := os.Mkdir(path, 0755)
+	if err != nil {
+		log.Warn("Failed to create directory: ", err)
```
carryall: @gutakk Why using `warn` instead of `error` here? Will the code below still works if the directory is not created?

workers/jobs/google_scraper.go
```diff
@@ -80,14 +79,14 @@ func (c *Context) PerformSearchJob(job *work.Job) error {
...
-func updateStatusToFailed(jobFails int64, jobName string, keywordID uint, keyword string, err error) {
+func updateStatusToFailed(jobFails int64, jobName string, keywordID uint, keyword string, failedReason error) {
 	if int(jobFails)+1 >= MaxFails {
-		updateStatusErr := google_search_service.UpdateKeywordStatus(keywordID, models.Failed, err)
+		err := google_search_service.UpdateKeywordStatus(keywordID, models.Failed, failedReason)
 
-		if updateStatusErr != nil {
-			panic(fmt.Sprintf("Cannot update keyword status (reason: %v)", updateStatusErr))
+		if err != nil {
+			log.Fatal("Failed to update keyword status: ", err)
```
carryall, line 87: @gutakk Should we add this as a constant in `helpers/error_handler/error_constant.go`?

workers/jobs/google_scraper.go
```diff
@@ -33,7 +32,7 @@ func (c *Context) PerformSearchJob(job *work.Job) error {
 	keywordID := uint(job.ArgInt64("keywordID"))
 	keyword := job.ArgString("keyword")
 	if keywordID == 0 {
-		log.Printf("Cannot perform job (reason: %v)", invalidKeywordIDError)
+		log.Error("Failed to perform job: ", invalidKeywordIDError)
```
carryall, line 35: @gutakk Should we add this as a constant in `helpers/error_handler/error_constant.go`?

services/keyword_service/keyword.go
gutakk: Good catch. Fixed.

helpers/session/session.go
gutakk: I decide to move duplicate logic to helper function in https://github.com/gutakk/go-google-scraper/issues/79.

Purpose of `error_constant.go` is share the message more than one file. I think after I finish https://github.com/gutakk/go-google-scraper/issues/79 task the `error_constant.go` is not necessary anymore.

In this case I think it is ok to leave this message here as it use only in this file.
What do you think?

workers/jobs/google_scraper.go
gutakk: Same as https://github.com/gutakk/go-google-scraper/pull/78#discussion_r573613521

workers/jobs/google_scraper.go
gutakk: Same as https://github.com/gutakk/go-google-scraper/pull/78#discussion_r573613521

gutakk:
@olivierobert
Move logrus to helper in 28e822a149f85675753af0a08401a4b96f51f74a
Move error constant to config in 1c268229ea4ddf1275b537b4d8d19b9fe696689a

I also add more info to PR description 🙏

helpers/error_handler/error_constant.go
gutakk: Fixed in 4057bb3668bf88d76ae2c862cbae07fc03e92ee8 06cf1b98470ef5fdeb40e1e1b3ff2f1b16a16826

services/google_search_service/request_test.go
gutakk: Good catch. I will do it in https://github.com/gutakk/go-google-scraper/issues/79

services/keyword_service/keyword.go
gutakk: Fixed in f8132d480a676f9c4ccab60430f3efbe5f3b9f8c

carryall (approved):

config/error/error_constant.go
```diff
@@ -0,0 +1,19 @@
+package error
+
+const (
+	ChangeToRootDirFailure   = "Failed to change to root directory: "
+	ConnectToDatabaseFailure = "Failed to connect database: "
+	DeleteRedisJobFailure    = "Failed to delete redis job: "
+	EnqueueJobFailure        = "Failed to enqueue job: "
+	HashPasswordFailure      = "Failed to hash password: "
```
olivierobert, line 8: While all the other error messages make sense in this file as they seem to apply to the whole application, I found it weird to have "HashPasswordFailure" as I thought this would be in a user model or service. 

However, I found out  `bcrypt.GenerateFromPassword` is used in many many places. 

image

Just as the dependency for log was imported in many places, there is the same issue with `golang.org/x/crypto/bcrypt`. Instead, there should be an internal package to abstract this and it should also host this error message to be specific to that package.

Not a blocker for this PR but this is an idea for further improvements. I also wanted to point it out so that you don't do the same mistake in the future. Do not sprinkle dependencies all over the place as this results in tight coupling and harder to debug/maintain applications.

olivierobert (approved):

config/error/error_constant.go
gutakk: Thanks for explanation.
I fixed this in https://github.com/gutakk/go-google-scraper/pull/80.

